PR #4, opened 2019-07-09
Add association between tasks and a Migration Plan

goosemania:
closes #5080
https://pulp.plan.io/issues/5080

### DIFF
--- a/pulp_2to3_migrate/app/viewsets.py
+++ b/pulp_2to3_migrate/app/viewsets.py
@@ -2,9 +2,11 @@ from drf_yasg.utils import swagger_auto_schema
 from rest_framework import mixins
 from rest_framework.decorators import detail_route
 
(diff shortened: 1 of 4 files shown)


pulp_2to3_migrate/app/viewsets.py
```diff
@@ -2,9 +2,11 @@
 from rest_framework import mixins
 from rest_framework.decorators import detail_route
 
+from pulpcore.app.viewsets.custom_filters import HyperlinkRelatedFilter
```
goosemania: Not available in the plugin API. Adding it there.

goosemania: https://github.com/pulp/pulpcore-plugin/pull/112

goosemania:
There will be an alternative solution for the problem - a task will report which resources were reserved.

PR #5, opened 2019-07-23
Save generic Pulp 2 content info for futher migration steps

goosemania:
It tries to save only newly added units, Pulp 2 content which is
already known to the migraiton tool is not re-queried or re-saved.

closes #5083
https://pulp.plan.io/issues/5083

### DIFF
--- a/pulp_2to3_migrate/app/constants.py
+++ b/pulp_2to3_migrate/app/constants.py
@@ -1,2 +1,35 @@
 # for tasking system to ensure only one migration is run at a time
-PULP_2TO3_MIGRATION_RESOURCE = "pulp_2to3_migration"
+PULP_2TO3_MIGRATION_RESOURCE = 'pulp_2to3_migration'
+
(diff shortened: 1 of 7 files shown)


pulp_2to3_migrate/app/constants.py
```diff
@@ -1,2 +1,35 @@
...
+    #     'Blob',
+    #     'Image',
+    #     'Manifest',
+    #     'ManifestList',
+    #     'Tag'],
+}
+
+# Mapping Pulp 2 content type to Pulp 3 plugin and content type
+# 'pulp2 content type id' -> ('pulp3 plugin', 'pulp3 content type')
+PULP_2TO3_TYPE_MAP = {
```
goosemania: Not used for now.
@dkliban , let me know if it's useful for the other task you wrote or I can remove it.

dkliban: It is useful. I need to look up model name from a string.

PR #9, opened 2019-08-19
Problem: Pulp2Content can't be queried using REST API

dkliban:
Solution: add REST API for Pulp2Content

This patch adds a ViewSet, a Filter, and a Serializer for Pulp2Content model. This enables REST API
users to retrieve a list of Pulp2Content as well as individual units. Lots of filters for filtering
are also available.

fixes: #5231
https://pulp.plan.io/issues/5231

### DIFF
--- a/pulp_2to3_migrate/app/serializers.py
+++ b/pulp_2to3_migrate/app/serializers.py
@@ -4,10 +4,11 @@ from rest_framework import serializers
 
 from pulpcore.plugin.serializers import (
     ModelSerializer,
+    DetailRelatedField,
     IdentityField
 )
 
-from .models import MigrationPlan
+from .models import MigrationPlan, Pulp2Content
 
 
 class MigrationPlanSerializer(ModelSerializer):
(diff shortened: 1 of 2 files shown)


goosemania:
@dkliban , I tested and it works as expected, thanks!

Do you think it makes sense to add pagination to this viewset?

Is the `pulp2content` a descriptive enough endpoint name or should we come up with a custom one but more self-explanatory?

dkliban:
@goosemania There is already pagination. Are you suggesting changing how pagination is presented to the user?

I don't have any other ideas for the endpoint name. I am open to suggestions though.

goosemania:
> @goosemania There is already pagination. Are you suggesting changing how pagination is presented to the user?

I see it now, I thought that one has to provide a paginator on a viewset explicitly.

> I don't have any other ideas for the endpoint name. I am open to suggestions though.

I think it's fine as it is, at least for now, however maybe "pulp-2to3-content-map" or something along those lines might be arguably clearer for user.

PR #11, opened 2019-08-25
Add models for pre0migrating pulp2 repository related data

goosemania:
closes #5306
https://pulp.plan.io/issues/5306

### DIFF
--- a/pulp_2to3_migrate/app/models/repository.py
+++ b/pulp_2to3_migrate/app/models/repository.py
@@ -1,0 +1,120 @@
+from django.contrib.postgres.fields import JSONField
+from django.db import models
+
+from pulpcore.app.models import (  # it has to be imported directly from pulpcore see #5353
+    Remote,
+    Publisher,
+)
+from pulpcore.plugin.models import (
+    Model,
+    BaseDistribution,
+    RepositoryVersion,
+    Publication,
+)
+
+
+class Pulp2Repository(Model):
+    """
+    Information about Pulp 2 repository.
+
+    Fields:
+        pulp2_repo_id (models.CharField): Repository ID in Pulp 2
+        pulp2_object_id (models.CharField): Object id of a repository in Pulp 2
+        pulp2_last_unit_added (models.DateTimeField): Last time a unit was added to
+            a repository in Pulp 2
+        pulp2_last_unit_removed (models.DateTimeField): Last time a unit was removed from
+            a repository in Pulp 2
+
+    Relations:
+        pulp3_repository_version (models.OneToOneField): Pulp 3 repository version which Pulp 2
+            repository was migrated to
+    """
+    pulp2_object_id = models.CharField(max_length=255, unique=True)
+    pulp2_repo_id = models.CharField(max_length=255)  # in Pulp2 it might be longer?
+    pulp2_last_unit_added = models.DateTimeField(null=True)
+    pulp2_last_unit_removed = models.DateTimeField(null=True)
+
+    pulp3_repository_version = models.OneToOneField(RepositoryVersion,
+                                                    on_delete=models.SET_NULL,
+                                                    null=True)
+
+
+class Pulp2RepoContent(Model):
+    """
+    Information about content in Pulp 2 repository.
+
+    It's important to use a relation to Pulp 2 repository and not just Pulp 2 repo_id because
+    over time repository with the same id can be removed and created again in Pulp 2 and
+    the migration plugin needs to distinguish between those.
+
+    Fields:
+        pulp2_unit_id (models.CharField): Repository id in Pulp 2
+        pulp2_content_type_id (models.CharField): Id of a content type in Pulp 2
+
+    Relations:
+        pulp2_repository (models.ForeignKey): Pulp 2 repository this content belongs to
+    """
+    pulp2_unit_id = models.CharField(max_length=255)
+    pulp2_content_type_id = models.CharField(max_length=255)
+
+    pulp2_repository = models.ForeignKey(Pulp2Repository, on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('pulp2_repository', 'pulp2_unit_id')
+
+
+class Pulp2Importer(Model):
+    """
+    Information about Pulp 2 importer.
+
+    Fields:
+       pulp2_type_id (models.CharField): Id of importer type in Pulp 2
+       pulp2_config (JSONField): Pulp 2 importer config in JSON format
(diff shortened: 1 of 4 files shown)


pulp_2to3_migrate/app/models/repository.py
```diff
@@ -0,0 +1,121 @@
...
+    pulp2_unit_id = models.CharField(max_length=255)
+    pulp2_content_type_id = models.CharField(max_length=255)
+
+    pulp2_repository = models.ForeignKey(Pulp2Repository, on_delete=models.CASCADE)
+
+    class Meta:
+        unique_together = ('pulp2_repository', 'pulp2_unit_id')
+
+
+class Pulp2Importer(Model):
```
goosemania: For this models migration generation fails like this 
```
    $ django-admin makemigrations
    Traceback (most recent call last):
      File "/usr/local/lib/pulp/bin/django-admin", line 11, in <module>
        sys.exit(execute_from_command_line())
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
        utility.execute()
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
        self.fetch_command(subcommand).run_from_argv(self.argv)
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/base.py", line 323, in run_from_argv
        self.execute(*args, **cmd_options)
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/base.py", line 361, in execute
        self.check()
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/base.py", line 390, in check
        include_deployment_checks=include_deployment_checks,
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/management/base.py", line 377, in _run_checks
        return checks.run_checks(**kwargs)
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/checks/registry.py", line 72, in run_checks
        new_errors = check(app_configs=app_configs)
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/checks/model_checks.py", line 168, in check_lazy_references
        return _check_lazy_references(apps)
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/checks/model_checks.py", line 163, in _check_lazy_references
        )), key=lambda error: error.msg)
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/checks/model_checks.py", line 162, in <genexpr>
        for func in apps._pending_operations[model_key]
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/checks/model_checks.py", line 157, in build_error
        return error_fn(model_key, func, args, keywords) if error_fn else None
      File "/usr/local/lib/pulp/lib64/python3.6/site-packages/django/core/checks/model_checks.py", line 107, in field_error
        'model': '.'.join(model_key),
    TypeError: sequence item 0: expected str instance, NoneType found
```
model_key is None for some reason

goosemania: It fails consistently but if *any* of the fields are commented out, migration generation succeeds from time to time :/

goosemania: I narrowed it down to Remote and Publisher.
If there is a relation to them, migration generation fails

goosemania: Solved by importing directly from pulpcore

pulp_2to3_migrate/app/models/repository.py
```diff
@@ -0,0 +1,123 @@
...
+        pulp2_last_unit_added (models.DateTimeField): Last time a unit was added to
+            a repository in Pulp 2
+        pulp2_last_unit_removed (models.DateTimeField): Last time a unit was removed from
+            a repository in Pulp 2
+
+    Relations:
+        pulp3_repository_version (models.OneToOneField): Pulp 3 repository version which Pulp 2
+            repository was migrated to
+    """
+    pulp2_repo_id = models.CharField(max_length=255)  # in Pulp2 it might be longer?
```
ipanova: can we use textfield then?wdyt?

pulp_2to3_migrate/app/models/repository.py
```diff
@@ -0,0 +1,123 @@
...
+    pulp2_object_id = models.CharField(max_length=255)
```
ipanova: what is the reason having both object_id and repo_id? can we refer only to repo_id?

pulp_2to3_migrate/app/models/repository.py
```diff
@@ -0,0 +1,123 @@
...
+    pulp2_last_unit_added = models.DateTimeField()
```
ipanova: what if this is an empty repo? can we set null=true for safety as well?

pulp_2to3_migrate/app/models/repository.py
goosemania: it can be but it's a part of a unique index now, so it's better be limited

pulp_2to3_migrate/app/models/repository.py
goosemania: Good question. Should we migrate empty repos? Maybe, why not. +1 then.

pulp_2to3_migrate/app/models/repository.py
goosemania: Since we can run a migration multiple times, it's possible that a repo with a specific repo_id can be removed and then created again and it will be a completely different repo. Migration plugin needs to distinguish between those.

pulp_2to3_migrate/app/models/repository.py
```diff
@@ -0,0 +1,123 @@
...
+    pulp2_last_unit_removed = models.DateTimeField(null=True)
+
+    pulp3_repository_version = models.OneToOneField(RepositoryVersion,
+                                                    on_delete=models.SET_NULL,
+                                                    null=True)
+
+    class Meta:
+        unique_together = ('pulp2_object_id', 'pulp2_repo_id')  # maybe just object_id?
```
ipanova: it seems like having object_id only would suffice

pulp_2to3_migrate/app/models/repository.py
goosemania: done

pulp_2to3_migrate/app/models/repository.py
goosemania: done

PR #23, opened 2019-09-25
Migrate on_demand content

goosemania:
Create RemoteArtifact for each of the lazy catalog entry.

closes #5337
https://pulp.plan.io/issues/5337


pulp_2to3_migration/app/models/content.py
```diff
@@ -74,3 +74,27 @@ def create_pulp3_content(self):
...
         raise NotImplementedError()
+
+
+class Pulp2LazyCatalog(Model):
+    """
+    Information for downloading Pulp 2 on_demand content.
+
+    Fields:
+        pulp2_importer_id (models.CharField): Importer ID in Pulp 2
+        pulp2_unit_id (models.CharField): Content unit ID in Pulp 2
```
dralley: Is this stored as a uuid in Pulp 2, or are mongodb IDs their own special thing?  (bson ID or something).

If UUID then we could store it as such.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -90,12 +98,22 @@ def __init__(self, migrator):
...
-    async def create_artifact(self, pulp2_storage_path, expected_digests={}, expected_size=None):
+    async def create_artifact(self, pulp2_storage_path, expected_digests={}, expected_size=None,
+                              downloaded=True):
         """
         Create a hard link if possible and then create an Artifact.
 
         If it's not possible to create a hard link, file is copied to the Pulp 3 storage.
+        For non-downloaded content, artifact with its expected checksum and size is created.
         """
+        if not downloaded:
```
dralley: If the unit has not been downloaded, then how can we create an artifact for it?  Should this be RemoteArtifact?

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -175,31 +193,71 @@ def __init__(self, migrator):
...
+
+            # Downloaded or on_demand content with LCEs.
+            #
+            # To create multiple remote artifacts, create multiple instances of declarative
+            # content which will differ by url/remote in their declarative artifacts
+            for lce in pulp2lazycatalog:
+                remote = get_remote_by_importer_id(lce.pulp2_importer_id)
+                deferred_download = not pulp2content.downloaded
+                if not remote and deferred_download:
+                    # on_demand content can't be migrated without a remote
```
dralley: Should this create a warning for the user?

pulp_2to3_migration/app/models/content.py
goosemania: @dralley, I think they are uuids. And on the content model I initially stored it like uuid. 
I ended up using just a char field to avoid conversion from uuid to string and back.
I don't remember where it was, probably if you search in mongo by this field. I can check or try to switch to it.
FWIW, this only used for mongo and uniquness constraints in pre-migrated models, this value is not migrated to Pulp 3.

pulp_2to3_migration/app/plugin/content.py
goosemania: It's an in-memory Artifact with digest and size (if provided) and nothing else in case of on_demand content. You need this object to pass to a DeclarativeArtifact, and later in the pipeline RemoteArtifacts will be created based on the provided info. Artifact objects won't be saved in this case.

pulp_2to3_migration/app/plugin/content.py
goosemania: Yeah, probably. +1

pulp_2to3_migration/app/plugin/content.py
dralley: Ah, they aren't saved.  +1

dralley:
Travis complaint ```Could not find changelog entry in CHANGES/ for 5337.```

Apart from having some sort of warning, LGTM

pulp_2to3_migration/app/models/content.py
dralley: It's unnecessary.  I was just thinking that a native uuid type would probably be a little faster and save some space in the DB/memory.  But not a big deal in either case, especially if this way has cleaner code.

PR #24, opened 2019-09-25
Add new changelog process

fabricio-aguiar:
- Adds a changelog update section to the docs
- creates the base CHANGES.rst file
- creates the CHANGES directory for news updates and indicates to git to
  store the dir even if empty with a .gitignore.
- adds a release note for this change
- adds a docs section that loads the CHANGES.rst file on the website

https://pulp.plan.io/issues/5501
re #5501


docs/changes.rst
```diff
@@ -0,0 +1,5 @@
+.. _pulp-2to3-migration-changes:
+
+.. include:: ../CHANGES.rst
+
+.. include:: ../HISTORY.rst
```
goosemania: I don't think we need a history file, since we haven't released anything yet (aka we don't have any history for release notes)

bmbouter: I think it's ok to leave here since a) the user won't see a difference since the empty HISTORY.rst is included on this page and b) this is consistent with all the other sites. If you decided to remove it I'm ok w/ that too.

docs/conf.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'pulp-2to3-migration'
+copyright = '2019, Pulp Team'
+author = 'Pulp Team'
+
+# The short X.Y version
+version = ''
```
bmbouter: I think you need to have this imported from the package. See this as an example: https://github.com/pulp/pulp_ansible/blob/master/docs/conf.py#L32-L35

docs/_templates/layout.html
```diff
@@ -0,0 +1,41 @@
...
+
+{% block footer %}
+{{ super() }}
+<script>
+  (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+  (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+  m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+  })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+
+  ga('create', 'UA-74748547-2', 'auto');
```
bmbouter: This is the property ID for docs.pulpproject.org, but I think GA will collate the data for other sites into it as well. I'm not 100% sure though.

docs/_templates/layout.html
```diff
@@ -0,0 +1,41 @@
...
+{# check current version against supported-releases.json #}
+<script>
+$( document ).ready( function() {
+
+  // gets version number minus whitespace and bugfix version
+  var $version_number = $('.version').text().trim().split('.', 2).join('.')
+
+  // div mirroring the admonition-warning generated by sphinx if added to .rst
+  var warning = '\
+          <div class="admonition warning"> \
```
bmbouter: To me, this admonition warning I think makes more sense on the main docs site (not a plugin site). Overall I think this whole file could be deleted.

bmbouter:
@fabricio-aguiar I'm done reviewing this PR. I left a comments for changes. I didn't formally LGTM, but if those changes are made I think it would be ready for merging.

docs/conf.py
```diff
@@ -0,0 +1,173 @@
...
+
+# -- Project information -----------------------------------------------------
+
+project = 'pulp-2to3-migration'
+copyright = '2019, Pulp Team'
+author = 'Pulp Team'
+
+
+# The short X.Y version
+version = pulp_2to3_migration.__version__
```
goosemania: @fabricio-aguiar, I don't think there is  __version__ defined in pulp_2to3_migration, please add it.
Have you tried to build docs with this config? It's interesting if it doesn't fail with the absent version.

bmbouter: @fabricio-aguiar I checked and I can see the version declared in the setup.py, but you also need to declare it again in this file  https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/__init__.py

See how other projects declare it again like here https://github.com/pulp/pulpcore/blob/master/pulpcore/__init__.py#L1

fabricio-aguiar: my bad, I haven't tried to build it before. It really failed, changed it and then I successfully built it

fabricio-aguiar: thanks @goosemania and @bmbouter

PR #28, opened 2019-10-03
Respect the migration plan

dralley:
closes #5450
https://pulp.plan.io/issues/5450
closes #5319
https://pulp.plan.io/issues/5319

### DIFF
--- a/pulp_2to3_migration/app/json_schema.py
+++ b/pulp_2to3_migration/app/json_schema.py
@@ -23,7 +23,7 @@ SCHEMA = '''{
                                 "name": {
                                     "type": "string"
                                 },
-                                "pulp2_repository_id": {
+                                "pulp2_importer_repository_id": {
                                     "type": "string"
                                 },
                                 "repository_versions": {
@@ -51,22 +51,9 @@ SCHEMA = '''{
                                     }
                                 }
                             },
-                            "required": ["name", "repository_versions"],
+                            "required": ["name", "pulp2_importer_repository_id", "repository_versions"],
                             "additionalProperties": false,
-                            "$comment": "pulp2_repository_id field should be specified so we know which importer to use when migrating multiple pulp2repos into repo versions",
-                            "if": {
-                            "properties": {
-                                    "repository_versions": {
-                                        "type": "array",
-                                        "minItems": 2
-                                    }
-                                }
-                            },
-                            "then": {
-                            "dependencies": {
-                                    "repository_versions": ["pulp2_repository_id"]
-                                }
-                             }
+                            "$comment": "pulp2_importer_repository_id field should be specified so we know which importer to use when migrating multiple pulp2 repos into repo versions"
                         }
                     }
                 },
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -17,14 +17,15 @@ from pulp_2to3_migration.app.plugin import PLUGIN_MIGRATORS
 _logger = logging.getLogger(__name__)
 
 
-async def migrate_content(plugins_to_migrate):
+async def migrate_content(plan):
     """
     A coroutine to initiate content migration for each plugin.
 
     Args:
-         plugins_to_migrate(list): List of plugins to migrate
+         plan (MigrationPlan): Migration Plan to use
     """
     content_migration_coros = []
+    plugins_to_migrate = plan.get_plugins()
 
     progress_data = dict(message='Migrating content to Pulp 3', code='migrating.content', total=0)
     with ProgressReport(**progress_data) as pb:
@@ -45,7 +46,7 @@ async def migrate_content(plugins_to_migrate):
         pb.done = pb.total
 
 
-async def migrate_repositories():
+async def migrate_repositories(plan):
     """
     A coroutine to migrate pre-migrated repositories.
     """
@@ -74,15 +75,17 @@ async def migrate_repositories():
                 pb.save()
 
 
-async def migrate_importers(plugins_to_migrate):
+async def migrate_importers(plan):
     """
     A coroutine to migrate pre-migrated importers.
 
     Args:
-        plugins_to_migrate(list): A list of plugins which are being migrated.
+        plan (MigrationPlan): Migration Plan to use.
     """
     # gather all needed plugin importer migrators
     importer_migrators = {}
+    plugins_to_migrate = plan.get_plugins()
+
     for plugin, plugin_migrator in PLUGIN_MIGRATORS.items():
         if plugin not in plugins_to_migrate:
             continue
--- a/pulp_2to3_migration/app/models/base.py
+++ b/pulp_2to3_migration/app/models/base.py
@@ -2,6 +2,13 @@ from django.contrib.postgres.fields import JSONField
 
 from pulpcore.plugin.models import Model
 
+from pulp_2to3_migration.pulp2 import connection
+from pulp_2to3_migration.pulp2.base import (
+    Distributor,
+    Importer,
+    Repository,
+)
+
 
 class MigrationPlan(Model):
     """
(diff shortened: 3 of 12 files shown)


dralley:
This is about 1/3 the code I had previously, because I realized my mental model of how it was supposed to work was wrong, and what I was doing was overcomplicated and incorrect.

My current understanding:

* The details of the migration plan is used only during pre-migration, for discovery of resources to migrate
* Pre-migration pulls them into PostgeSQL
* The migration steps migrate *everything* found by pre-migration, and have no further use for the "plan" except as a list of plugins (this is what I think I misunderstood earlier, I was trying to change the migration steps when that was unnecessary)

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +17,105 @@ class MigrationPlan(Model):
...
+
+    def validate_and_populate(self):
+        for plugin_data in self.migration_plan['plugins']:
+            self.plugins_to_migrate.append(plugin_data['type'])
+            if plugin_data.get('repositories'):
+                self._parse_repository_data(plugin_data.get('repositories'))
+            # protection
+
+    def _parse_repository_data(self, repository_data):
+        for repository in repository_data:
```
dralley: I'm not doing anything with  "name" here - I'm still not sure what it is for exactly.  Is it the name of the Pulp 3 repository to be created?  Or does it relate to Pulp 2?

dralley:
Right now I know there's some problems around the handling of "repo_id" keys causing importers not to be migrated properly

pulp_2to3_migration/app/tasks/migrate.py
```diff
@@ -28,18 +28,15 @@ def migrate_from_pulp2(migration_plan_pk, dry_run=False):
         migration_plan_pk (str): The migration plan PK.
         dry_run (bool): If True, nothing is migrated, only validation happens.
     """
+    # MongoDB connection initialization
+    connection.initialize()
+
     plan = MigrationPlan.objects.get(pk=migration_plan_pk)
+
```
ipanova: do we still want to keep dry-run mode? right now it does have any impact other than a logger message

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +17,113 @@ class MigrationPlan(Model):
...
+        self.importers_to_migrate = []
+        self.distributors_to_migrate = []
+        self.repositories_to_migrate = []
+
+        self.validate_and_populate()
+
+    def validate_and_populate(self):
+        for plugin_data in self.migration_plan['plugins']:
+            self.plugins_to_migrate.append(plugin_data['type'])
+            if plugin_data.get('repositories'):
```
ipanova: @goosemania given that fact that we changed our migration strategy( i.e. we cannot migrate only the content because we need data from repo/importers),  i think we need to make repository as required field on the schema of the MP

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +17,113 @@ class MigrationPlan(Model):
...
+            except Repository.DoesNotExist:
+                raise SchemaValidationFailure(
+                    "Repository id '{}' specified does not exist in Pulp2".format(
+                        repository['pulp2_repository_id']
+                    ))
+
+            try:
+                importer = Importer.objects.get(repo_id=repo.repo_id)
+                self.importers_to_migrate.append(importer.id)
+            except Importer.DoesNotExist:
```
ipanova: i believe we should raise an error here as well, @dralley ?

pulp_2to3_migration/app/models/base.py
goosemania: @ipanova, either this or in the past we agreed that we just skip repo/importer parts and migrate only downloaded content. I'l create a task  which will either change a requirement or skip unnecessary steps.

pulp_2to3_migration/app/tasks/migrate.py
dralley: I think we can remove it for now and decide if we want it later.  It would be a struggle I think because the migrate steps work on the data created by the pre-migrate steps so if we're not saving any of that data, we can't know what the migrate steps are supposed to be copying.

We could have a dry-run mode that skips the mgiration steps and just logs what is pre-migrated.  But that can be discussed further I guess.

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +17,113 @@ class MigrationPlan(Model):
...
+        """
+        return self.plan_view.plugins_to_migrate
 
     def get_repositories(self):
         """
         Return a list of pulp2 repositories to migrate or empty list if all should be migrated.
         """
-        # TODO: get pulp2 repositories from the self.plan
-        return []
+        return self.plan_view.repositories_to_migrate
```
goosemania: @dralley , I thought we decided on IRC that to create repo versions we need a structure for repositories and that just a list of ids is not enough.
For importers/distributors, it's sufficient, however for repositories we need all the info to create a repo version, to create a distribution and so on. 
E.g. repositories can be a dict with repo name as a key.

Plan:
```
"repositories":
[
           {
             "name": "foo",
             "pulp2_repository_id":  "my_repo_id to use to find an importer to migrate"   # optional?
             "repository_versions": [
                 {
                   "pulp2_repository_id": "idA",
                   "distributor_ids":  ["distributor_id1", "distributor_id2"]  # optional?
                   "protection": ["distributor_id1"] # content protection migration is only needed for the specfied distributor(s)
                  },                
                  {
                    "pulp2_repository_id": "idB",
                    "distributor_ids":  ["distributor_id3"]
                  },
               ]
            }
]
```

get_repositories returns a dict with a repo_id as a key:
```
{ "foo": 
      {"importer_from_repo":  "my_repo_id to use to find an importer to migrate",
       "repo_versions": [  as they are in migration plan ]
      }
}
```

What do you think?

goosemania:
@dralley , it's nice that you added validation, thank you!

I think we ought to make it optional. Depending on the size of a migration plan there can be a lot of calls to mongo and it can be potentially slow.
Another issue is that most of the time users will run migration while Pulp 2 is running, so it's possible that a migration plan is not fully valid and it's fine. All the migration code should not rely on all resources being present in pulp2 at the time of pre-migration.

I see the following goals of validation:
     - the structure is right
     - the majority of resources exist in Pulp 2 if Pulp 2 is running
     - all resources exist in Pulp 2 if Pulp 2 is not running

The last two cases are assessed by the user, validator has no info if pulp 2 is running or not.
So it would be nice if validation can accumulate invalid/non-existing resources to report to user. 

Any thoughts?

pulp_2to3_migration/app/models/base.py
dralley: Since these importers are found only by discovery via the listed repositories, I didn't think that was necessary.  Unlike the Repository and Distributor IDs they are not actually specified by the user in the sync plan.

pulp_2to3_migration/app/models/base.py
dralley: @goosemania Please link me the task when you create it

dralley:
@goosemania That makes sense.  So the plan is to provide a validation parameter on the POST request to run the migration plan?

I'm not sure I'm worried about the performance so much, but maybe I'm lacking context of the scale of the potential problem.  My thought is that even if there are thousands of repos in the migration plan, the database traffic would absolutely pale in comparison to (pre)migrating content, and the simplicity of doing all the migration plan processing in one place would be worth it.  Let me know if my assumptions are off, there.  But, we could still improve the queries by making lists of resources and then doing a couple of larger queries instead of a bunch of smaller ones.

It makes sense to accumulate the missing entries for reporting purposes instead of immediately throwing a ValidationError, I will work that into this code.

pulp_2to3_migration/app/models/base.py
ipanova: in case this we are using importer of provided pulp2_repository_id, which will transferred into remote in pulp3 terms.
it is true that it is quite unlikely to find a repo without importer in pulp2, but in case we do, it will prevent us from further remote creation, as @goosemania  mentioned, it would make sense to accumulate missing entries.

pulp_2to3_migration/app/models/base.py
goosemania: @dralley https://pulp.plan.io/issues/5549

dralley:
@ipanova @goosemania I didn't get a chance to try these new changes so I'm 100% sure it doesn't work, but you can re-review the basic flow.

pulp_2to3_migration/pulp2/connection.py
```diff
@@ -53,12 +53,8 @@ def initialize(name=None, seeds=None, max_pool_size=None, replica_set=None, max_
 
     # We do not allow a second call to initialize(), as mongoengine.connect() will cache the last
     # initialized connection for all calls. Thus, any process that attempts to call initialize()
-    # again might alter which database all further queries are made against. By raising this
-    # Exception, we can ensure that only one database connection is established per process which
-    # will help us to ensure that the connection does not get overridden later.
+    # again might alter which database all further queries are made against.
```
dralley: There's no reason to warn with the guards in place anyways, and we may need to sprinkle this in more than one place to ensure it will happen on all codepaths.

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +19,136 @@ class MigrationPlan(Model):
...
+        for repository in repository_data:
+            name = repository_data['name']
+
+            _find_importer_repo = repository['pulp2_repository_id']
+            self.importer_discovery_repos.append(_find_importer_repo)
+
+            repository_versions = [] or \
+                self._parse_repository_version_data(repository.get('repository_versions'))
+
+            repository_data = {
```
ipanova: i think then intention was to include _find_importer_repo inside of this structure. This way we will know which importer to use when creating remote for repo named X form this repo_data structure. did you have something else on your mind?

goosemania: +1 to add importer info, we need to relate it later. After migrating everyting to Pulp3, user needs to know which Remote to use with which Repository. And this information is available here only.

It would also be convenient to have a dictionary here instead of a list, because repositories are created first and then they are getting migrated so getting information about repo versions by `name` would be nice.

pulp_2to3_migration/app/json_schema.py
```diff
@@ -51,9 +51,9 @@
                                     }
                                 }
                             },
-                            "required": ["name", "repository_versions"],
+                            "required": ["name", "importer_repository_id", "repository_versions"],
                             "additionalProperties": false,
-                            "$comment": "pulp2_repository_id field should be specified so we know which importer to use when migrating multiple pulp2repos into repo versions",
+                            "$comment": "importer_repository_id field should be specified so we know which importer to use when migrating multiple pulp2repos into repo versions",
                             "if": {
                             "properties": {
```
dralley: Does line 67 need to be changed?  it looks like it's referring to the nested one.  But if that's the case then we might be able to get rid of the conditional because it's already required on line 45?

goosemania: yeah, I think all conditional can go away. @ipanova , should confirm :)

ipanova: i confirm ^_^

ipanova: you can remove if/then conditionals

pulp_2to3_migration/app/migration.py
```diff
@@ -17,14 +17,15 @@
 _logger = logging.getLogger(__name__)
 
 
-async def migrate_content(plugins_to_migrate):
```
dralley: It'll be better to just pass in the plan everywhere for consistency, and then the functions can use whatever parts of the plan they need.  The info is pre-processed and cached so there is no cost to doing so.

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +19,164 @@ class MigrationPlan(Model):
...
         """
-        # TODO: get pulp2 distributors from the self.plan
-        return []
+        return self.plan_view.distributors_to_migrate
+
+    def get_missing_resources(self):
+        """
+        Return a dict of any resources listed in the plan but missing from Pulp 2.
+
+        Repositories and Importers are enumerated by repo_id, Distributors by distributor_id.
```
dralley: Unless it would make sense to do distributors by distributor_id also.

pulp_2to3_migration/app/models/__init__.py
```diff
@@ -10,3 +10,12 @@
     Pulp2RepoContent,
     Pulp2Repository,
 )
+
+
```
dralley: Might be a better place for this, I just stuck it somewhere.

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +19,147 @@ class MigrationPlan(Model):
...
 
     def get_repositories(self):
         """
         Return a list of pulp2 repositories to migrate or empty list if all should be migrated.
         """
-        # TODO: get pulp2 repositories from the self.plan
-        return []
+        return self.plan_view.repositories_to_migrate
+
+    def get_pulp3_repository_setup(self):
```
dralley: This is the verbose version that returns data structures

pulp_2to3_migration/app/models/base.py
```diff
@@ -11,24 +19,147 @@ class MigrationPlan(Model):
...
+            pulp2_repository_id = repository_version['pulp2_repository_id']
+            self.repositories_to_migrate.append(pulp2_repository_id)
+            repository_versions.append(pulp2_repository_id)
+
+            distributor_ids = repository_version.get('distributor_ids')
+            self.distributors_to_migrate.extend(distributor_ids)
+
+        return repository_versions
+
+    def _check_missing(self):
```
dralley: Notably this does not modify the existing representations of repositories, importers, distributors to migrate.  Should it?

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -292,7 +293,7 @@
 
     # in case only certain importers are specified in the migration plan
     if importers:
-        mongo_importer_q &= mongo_Q(pulp2_id__in=importers)
+        mongo_importer_q &= mongo_Q(repo_id__in=importers)
```
dralley: ```pulp2_id``` doesn't exist

pulp_2to3_migration/app/serializers.py
```diff
@@ -15,7 +15,7 @@
     IdentityField
 )
 
-from .json_schema import SCHEMA
+from pulp_2to3_migration.app import json_schema
```
dralley: Part of a refactor I gave up on -- I can undo this if you like.

dralley:
Not sure what's up with the Travis error.

pulp_2to3_migration/app/models/base.py
goosemania: Did you mean by repo_id? I think it's fine as it is for now. We might change something when we migrate distributors.

pulp_2to3_migration/app/models/__init__.py
dralley: Probably, I missed that file somehow

pulp_2to3_migration/app/models/base.py
dralley: Yup, that's what I meant.  Thanks

PR #31, opened 2019-10-08
Don't migrate importers/distributors with an empty config

goosemania:
There is nothing to migrate to Pulp3 in this case.

closes #5551
https://pulp.plan.io/issues/5551


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -286,14 +286,19 @@
                          all are migrated.
     """
     mongo_importer_q = mongo_Q(repo_id=repo.pulp2_repo_id)
+
+    # importers with empty config are not needed - nothing to migrate
+    mongo_importer_q &= mongo_Q(config__ne={})
```
ipanova: it might worth adding a check if feed is null, wdyt? remotes in pulp3 are tightly coupled with the feed

PR #32, opened 2019-10-09
Handle the case when importer has no config at all

goosemania:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -288,7 +288,7 @@ async def pre_migrate_importer(repo, importers):
     mongo_importer_q = mongo_Q(repo_id=repo.pulp2_repo_id)
 
     # importers with empty config are not needed - nothing to migrate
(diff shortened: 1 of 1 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -288,7 +288,7 @@
     mongo_importer_q = mongo_Q(repo_id=repo.pulp2_repo_id)
 
     # importers with empty config are not needed - nothing to migrate
-    mongo_importer_q &= mongo_Q(config__ne={})
+    mongo_importer_q &= mongo_Q(config__ne={}) & mongo_Q(config__exists=True)
```
ipanova: tiny suggestion to switch the order, if config does not exist, second part wont be executed

goosemania: really? I thoughy mongo doesn't have lazy evaluation.
I'll change, just in case.

goosemania:
Permission failures in travis are unrelated to the changes of this PR. Merging to unblock Katello.

PR #36, opened 2019-10-17
Fix "migration plan needs to be pre-serialized"

dralley:
closes: #5546
https://pulp.plan.io/issues/5546

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -53,7 +53,7 @@ diagrams:
 
 html:
 	mkdir -p $(BUILDDIR)/html
-	curl -o _build/html/api.json "http://localhost:24817/pulp/api/v3/docs/?format=openapi"
+	curl -o _build/html/api.json "http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_2to3_migration"
 	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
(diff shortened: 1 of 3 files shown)


docs/Makefile
```diff
@@ -53,7 +53,7 @@ diagrams:
 
 html:
 	mkdir -p $(BUILDDIR)/html
-	curl -o _build/html/api.json "http://localhost:24817/pulp/api/v3/docs/?format=openapi"
+	curl -o _build/html/api.json "http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_2to3_migration"
```
dralley: Mirror all the other plugin docs

dralley:
I was able to test this and it works

pulp_2to3_migration/app/serializers.py
```diff
@@ -43,7 +43,10 @@ def validate(self, data):
         """
         schema = json.loads(SCHEMA)
         validator = Draft7Validator(schema)
-        loaded_plan = json.loads(data['plan'])
+        if isinstance(data['plan'], str):
+            loaded_plan = json.loads(data['plan'])
+        elif isinstance(data['plan'], dict):
```
goosemania: Does it make sense to cover the `else`?
Just in case someone provided `list` by mistake, which is still a valid json. What do you think?

PR #39, opened 2019-10-18
Create repo versions

goosemania:
By default, create one repo version per each pulp 2 repo.
If the specific migration plan exists for repo version creation, follow it.

Repositories are created based on the provided name in the migration plan.
If such repo exists in pulp 2, its description will be migrated, otherwise
Pulp 3 repo is created with the descritpion as its name.

closes #5342
https://pulp.plan.io/issues/5342

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -3,13 +3,16 @@ import hashlib
 import logging
 
 from pulpcore.plugin.models import (
+    Content,
     ProgressReport,
     Repository,
+    RepositoryVersion,
 )
 
 from pulp_2to3_migration.app.models import (
     Pulp2Content,
     Pulp2Importer,
+    Pulp2RepoContent,
     Pulp2Repository,
 )
 from pulp_2to3_migration.app.plugin import PLUGIN_MIGRATORS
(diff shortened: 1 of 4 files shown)


pulp_2to3_migration/app/migration.py
```diff
@@ -50,29 +53,56 @@
...
+            for pulp3_repo_name in repos_to_create:
+                try:
+                    pulp2repo = pulp2repos_qs.get(pulp2_repo_id=pulp3_repo_name)
+                except Pulp2Repository.DoesNotExist:
+                    description = pulp3_repo_name
+                else:
+                    description = pulp2repo.pulp2_description
+
+                repo, created = Repository.objects.get_or_create(
+                    name=pulp3_repo_name[:255],
```
dralley: Why use a CharField here instead of TextField?  We could ditch the 255 character limitation.

pulp_2to3_migration/app/migration.py
```diff
@@ -50,29 +53,56 @@
...
+                    pb.save()
+
+        # specific migration plan for repositories
+        else:
+            pb.total += len(repos_to_create)
+            pb.save()
+
+            for pulp3_repo_name in repos_to_create:
+                try:
+                    pulp2repo = pulp2repos_qs.get(pulp2_repo_id=pulp3_repo_name)
```
dralley: >(pulp2_repo_id=pulp3_repo_name)

This looks a little like a typo but I'm not sure -- why is the Pulp 2 repo being looked up by a Pulp 3 repo name?

pulp_2to3_migration/app/migration.py
goosemania: We have this in the core https://github.com/pulp/pulpcore/blob/master/pulpcore/app/models/repository.py#L45 and I think the limitation is mainly for index purposes.

pulp_2to3_migration/app/migration.py
goosemania: The idea is that if you preserve pulp2 name (you set in the MP to create a pulp3 repo with the same name) you likely want to move a repo as is, including its description. Maybe its a wrong assumption, I'm open for feedback.

pulp_2to3_migration/app/migration.py
dralley: I'm pretty sure that's just a leftover from when we wanted to support MariaDB, which can't index TextFields so you have to use CharField.  PostgreSQL can, though, so we can use either.  https://docs.djangoproject.com/en/2.2/ref/databases/#indexes-for-varchar-and-text-columns

pulp_2to3_migration/app/migration.py
dralley: Ack, it's fine, I just wanted to be sure there was a reason.

pulp_2to3_migration/app/migration.py
dralley: We should change this in core too.  We already did in a lot of places, looks like this one got missed.

PR #44, opened 2019-10-31
New endpoint to query the Pulp2-Pulp3 mapping

fabricio-aguiar:
https://pulp.plan.io/issues/5634
closes #5634

### DIFF
--- a/pulp_2to3_migration/app/viewsets.py
+++ b/pulp_2to3_migration/app/viewsets.py
@@ -18,11 +18,12 @@ from pulpcore.plugin.viewsets import (
 )
 
 from .constants import PULP_2TO3_MIGRATION_RESOURCE
-from .models import MigrationPlan, Pulp2Content
+from .models import MigrationPlan, Pulp2Content, Pulp2Repository
 from .serializers import (
     MigrationPlanSerializer,
     MigrationPlanRunSerializer,
-    Pulp2ContentSerializer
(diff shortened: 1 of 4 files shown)


pulp_2to3_migration/app/viewsets.py
```diff
@@ -94,3 +95,26 @@ class Pulp2ContentViewSet(NamedModelViewSet, mixins.RetrieveModelMixin, mixins.L
...
+        fields = {
+            'pulp2_repo_id': ['exact', 'in'],
+        }
+
+
+class Pulp2to3MappingViewSet(NamedModelViewSet, mixins.RetrieveModelMixin, mixins.ListModelMixin):
+    """
+    ViewSet for Pulp2to3Mapping model.
+    """
+    endpoint_name = 'pulp2to3mapping'
```
dkliban: Let's change this endpoint name to 'pulp2repositories'.

Maybe also change the class names for the viewset and serializer?

fabricio-aguiar: @goosemania wdyt?

ipanova: +1

ipanova:
@fabricio-aguiar can you add the new endpoint to the user docs? https://github.com/pulp/pulp-2to3-migration/blob/master/README.md#user-guide

PR #49, opened 2019-11-22
Adds functional test that migrates 4 ISO repos

dkliban:
Install MongoDB on Travis
Build bindings for pulpcore and the migration plugin
Clone https://github.com/pulp/pulp-2to3-migration-test-fixtures
Populate MongoDB and /var/lib/pulp using the fixtures repo
Run a test that creates a migration plan and runs it.

### DIFF
--- a/pulp_2to3_migration/tests/functional/test_file_migration.py
+++ b/pulp_2to3_migration/tests/functional/test_file_migration.py
@@ -1,9 +1,70 @@
+from time import sleep
 import unittest
+from pulpcore.client.pulpcore import (ApiClient as CoreApiClient, Configuration,
+                                      TasksApi)
+from pulpcore.client.pulp_2to3_migration import (ApiClient as MigrationApiClient,
+                                                 MigrationPlansApi)
+
+
+def monitor_task(tasks_api, task_href):
+    """Polls the Task API until the task is in a completed state.
+
+    Prints the task details and a success or failure message. Exits on failure.
+
+    Args:
+        tasks_api (pulpcore.client.pulpcore.TasksApi): an instance of a configured TasksApi client
+        task_href(str): The href of the task to monitor
+
+    Returns:
+        list[str]: List of hrefs that identify resource created by the task
+
+    """
+    completed = ['completed', 'failed', 'canceled']
+    task = tasks_api.read(task_href)
+    while task.state not in completed:
+        sleep(2)
+        task = tasks_api.read(task_href)
+    if task.state == 'completed':
+        print("The task was successfful.")
+    else:
+        print("The task did not finish successfully.")
+    return task
 
 
 class TestMigrationPlan(unittest.TestCase):
     """Test the APIs for creating a Migration Plan."""
 
+    @classmethod
+    def setUpClass(cls):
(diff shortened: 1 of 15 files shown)


pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -0,0 +1,99 @@
+import socket
+from time import sleep
+import unittest
+from pulpcore.client.pulpcore import (ApiClient as CoreApiClient, Configuration,
```
goosemania: Could we have every item on the new line? Similar to the style we have in pulpcore or pulp_rpm?

```
from ... import (
    module1,
    module2,
)
```

pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -0,0 +1,99 @@
...
+        configuration.host = 'http://{}:24817'.format(socket.gethostname())
+        configuration.safe_chars_for_path_param = '/'
+
+        core_client = CoreApiClient(configuration)
+        file_client = FileApiClient(configuration)
+        migration_client = MigrationApiClient(configuration)
+
+        # Create api clients for all resource types
+        # artifacts = ArtifactsApi(core_client)
+        cls.file_repositories = RepositoriesFileApi(file_client)
```
goosemania: I'm not sure about naming, because it sounds like we store objects here, like file repository instances. maybe API should be there? e.g. `cls.file_repo_api` ?

pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -0,0 +1,99 @@
...
+        cls.file_content = ContentFilesApi(file_client)
+        # filedistributions = DistributionsFileApi(core_client)
+        # filepublications = PublicationsFileApi(file_client)
+        # fileremotes = RemotesFileApi(file_client)
+        cls.tasks = TasksApi(core_client)
+        # uploads = UploadsApi(core_client)
+        cls.migration_plans = MigrationPlansApi(migration_client)
+
+    def test_migrate_iso_repositories(self):
+        """Test that a valid Migration Plan can be created."""
```
goosemania: It seems like test doesn't run MP validation but it checks few things for the results of the migraiton. 
So I'm not sure if this docsting is up to date.

pulp_2to3_migration/tests/functional/test_file_migration.py
dkliban: yes! I will do that.

pulp_2to3_migration/tests/functional/test_file_migration.py
dkliban: yes! i renamed all to have _api on the end of the name.

pulp_2to3_migration/tests/functional/test_file_migration.py
dkliban: Yes, it's definitely out of date.

dralley:
When I run this locally, I get:

Screenshot from 2020-01-02 12-46-37

PR #66, opened 2020-01-02
Refactor migration plan parsing to keep track of plugin types

dralley:
closes: #5845
https://pulp.plan.io/issues/5845
closes: #5899
https://pulp.plan.io/issues/5899

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -14,7 +14,6 @@ from pulp_2to3_migration.app.models import (
     Pulp2RepoContent,
     Pulp2Repository,
 )
-from pulp_2to3_migration.app.plugin import PLUGIN_MIGRATORS
 
 _logger = logging.getLogger(__name__)
 
@@ -27,17 +26,15 @@ async def migrate_content(plan):
          plan (MigrationPlan): Migration Plan to use
     """
     content_migration_coros = []
-    plugins_to_migrate = plan.get_plugins()
 
     progress_data = dict(message='Migrating content to Pulp 3', code='migrating.content', total=0)
     with ProgressReport(**progress_data) as pb:
         # schedule content migration into Pulp 3 using pre-migrated Pulp 2 content
-        for plugin in plugins_to_migrate:
-            plugin_migrator = PLUGIN_MIGRATORS.get(plugin)
-            content_migration_coros.append(plugin_migrator.migrate_content_to_pulp3())
+        for plugin in plan.get_plugin_plans():
+            content_migration_coros.append(plugin.migrator.migrate_content_to_pulp3())
 
             # only used for progress bar counters
-            content_types = plugin_migrator.content_models.keys()
+            content_types = plugin.migrator.content_models.keys()
             pulp2content_qs = Pulp2Content.objects.filter(pulp2_content_type_id__in=content_types,
                                                           pulp3_content=None)
             pb.total += pulp2content_qs.count()
@@ -53,7 +50,6 @@ async def migrate_repositories(plan):
     A coroutine to migrate pre-migrated repositories.
     """
 
-    repos_to_create = plan.get_pulp3_repository_setup().keys()
     progress_data = dict(
         message='Creating repositories in Pulp 3', code='creating.repositories', total=0
     )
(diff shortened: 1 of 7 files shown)


dralley:
@goosemania @ipanova  -- Justin and I ended up getting blocked on #5899 during the week before break, so I interrupted work on the other issues to work on it.  It seemed like the most straightforwards way to solve it was to implement #5845, so I did that too.  This is a big refactor and (I think) a good improvement to  migration plan usage, but I haven't tested it thoroughly yet and it might still break things.  Let me know what you think.

dralley:
From my comment on the issue

> 
> In my PR I ended up doing something slightly different that I think will end up being much cleaner.  Instead of getting a single "pulp3_repo_setup" structure which is discriminated by type, I modified all of the parsing code so that it creates many "plugin migration plans" which you acquire using the method "plan.get_plugin_plans()".
> 
> Each of those plugin migration plan objects keep track of all their {repositories to create/repositories/importers/distributors} separately instead of merging them all into one big list (although I kept the methods to get the complete lists).  I noticed that we were looking up the "migrators" a lot, so the new object also makes the associated "migrator" (if it exists) an attribute, as well as whether the plan is "empty".  That helped to clean up the code a lot and will hopefully allow it to be cleaned up more later.

pulp_2to3_migration/app/migration.py
```diff
@@ -206,22 +187,32 @@
...
+            pulp3_repo_setup = plugin.get_repo_creation_setup()
+            if not pulp3_repo_setup:
+                for pulp2dist in pulp2distributors_qs:
+                    dist_migrator = distributor_migrators.get(pulp2dist.pulp2_type_id)
+                    await migrate_repo_distributor(pb, dist_migrator, pulp2dist)
+            else:
+                for repo_name in pulp3_repo_setup:
+                    for repo_dist in pulp3_repo_setup[repo_name]['versions']:
+                        pass
+                        # find pulp2repo by id
```
dralley: This code is broken, but I think it is also broken _currently_, I just never reach it on master branch because of https://pulp.plan.io/issues/5899

What currently gets returned by ```plan.get_pulp3_repo_setup()``` (both on master and in this PR) looks like [0] which isn't what this code expects I think.

[0]https://github.com/pulp/pulp-2to3-migration/pull/66/files#diff-2e8369e7a70124f8b4f1bc6f4120718eR171-R180 

I think it's supposed to be doing something more like this, correctt?

```
migrated_repo = Pulp2Repository.objects.get(pulp2_repo_id=repo_dist)
repo_distributors = Pulp2Distributor.objects.filter(pulp2_repository_id=migrated_repo.pk)
for pulp2dist in repo_distributors:
dist_migrator = distributor_migrators.get(pulp2dist.pulp2_type_id)
await migrate_repo_distributor(
     pb, dist_migrator, pulp2dist, migrated_repo.pulp3_repository_version)
```

pulp_2to3_migration/app/migration.py
```diff
@@ -120,12 +111,11 @@
     """
     # gather all needed plugin importer migrators
     importer_migrators = {}
-    plugins_to_migrate = plan.get_plugins()
 
-    for plugin, plugin_migrator in PLUGIN_MIGRATORS.items():
-        if plugin not in plugins_to_migrate:
+    for plugin in plan.get_plugin_plans():
```
dralley: This pattern is common enough that it might make sense to have a ```with_migrator=True``` kwarg on this method that just filters out plugins that don't include migrators, agree?

dralley:
I tested the new smash tests locally and even using master branch I get the same error that the CI returned.  It seems to be a problem with the bindings?

pulp_2to3_migration/app/models/base.py
```diff
@@ -68,93 +70,143 @@ def get_missing_resources(self):
...
-        if self.plan_view.missing_importers:
-            ret['importers'] = self.plan_view.missing_importers
-        if self.plan_view.missing_distributors:
-            ret['distributors'] = self.plan_view.missing_distributors
+            ret['repositories'] = self.plan_view.repositories_missing_repositories
+        if self.plan_view.repositories_missing_importers:
+            ret['repositories_missing_importers'] = self.plan_view.missing_importers
+        if self.plan_view.repositories_missing_distributors:
+            ret['repositories_missing_distributors'] = \
+                self.plan_view.repositories_missing_distributors
```
dralley: Does it make sense to make the ```missing_resources``` plugin specific as well?

pulp_2to3_migration/app/models/base.py
```diff
@@ -68,93 +70,143 @@ def get_missing_resources(self):
...
-
-    def _parse_repository_version_data(self, repository_version_data):
-        repository_versions = []
-
-        for repository_version in repository_version_data:
-            pulp2_repository_id = repository_version['pulp2_repository_id']
-            self.repositories_to_migrate.append(pulp2_repository_id)
-            repository_versions.append(pulp2_repository_id)
+    @property
+    def all_repositories_importers_to_migrate(self):
```
dralley: Does it make sense to move towards just doing everything on a plugin-by-plugin basis and not having these huge lists of all the repos/importers/distributors of various types mixed together?

pulp_2to3_migration/app/migration.py
goosemania: I'm not sure when we can do anything without a migrator for a plugin being present, so it might worth returning only plugins with a migrator.

pulp_2to3_migration/app/models/base.py
goosemania: If it's more convenient to do so because you refactor to have plans by plugin, sure, go for it.
Otherwise, as I usually do, I'd postpone until it's needed.

pulp_2to3_migration/app/models/base.py
goosemania: I think we pre-migrate repositories all together since they are not typed per se in pulp2.
It's convenient to check if repository importer or distributor needs to be migrated when we have a list of all of them and not by type.
Having said that, we can chain them here or whenever we need them. So it's not a problem to switch to plugin-by-plugin lists.

pulp_2to3_migration/app/migration.py
goosemania: I think the code expects this https://pulp.plan.io/issues/5837#Example-of-parsed-plan.
I'd defer it to @ipanova since she worked on it.

goosemania: I think the code expects this https://pulp.plan.io/issues/5837#Example-of-parsed-plan.
I'd defer it to @ipanova since she worked on it.

pulp_2to3_migration/app/models/base.py
dralley: I don't see an urgent need so we can just postpone, but I didn't know if it might be useful.

pulp_2to3_migration/app/migration.py
dralley: Awesome, I will make that change.  I noticed one place where ```get_plugin_plans()``` was being used without doing that check, but I just realized that the rest of the code is using the migrator so that check should be there but is missing.

pulp_2to3_migration/app/models/base.py
dralley: That makes sense, we can just keep these methods here then.

pulp_2to3_migration/app/migration.py
dralley: Eventually yes (I'm working on that issue) but right now it isn't getting data formatted like that.  I'm not sure if I should just leave it commented out for now (until that patch is finished) or if I should add a workaround here.

goosemania: I think @ipanova imlpied that it would work with the 5837 implemented only.
The workaround you suggested won't work in case distributor(s) should be used from a different repo. I think this is the main reason we need 5837.

dralley: Then I'll just leave it commented out until 5837

dralley:
@goosemania So I figured out what the problem is, I think, and I'm *really* confused.  Like I mentioned this happens to me on master branch as well.  All this testing below is from the master branch.

In this screenshot, I just ran a ```pclean``` and then subsequently ```pytest -v -r sx --color=yes --pyargs pulp_2to3_migration.tests.functional``` so that it should be doing the migration.  The test itself creates a migration plan with just "iso" specified and not doing any explicit naming.  Everything gets pre-migrated and the repos get migrated, as well as remotes, but a new repo version with the content relationships isn't being created.

Screenshot from 2020-01-06 23-28-55

Error message

Screenshot from 2020-01-06 23-29-53

I dug into the code and see that the set is empty because none of the Pulp2Content objects have ```pulp3_object``` set, so ```pulp3_content__pk``` resolves to ```None``` and the set math is all messed up.

Screenshot from 2020-01-06 23-38-53

And then I follow the chain further and that's where it gets really weird.

I get to this code here which is what is supposed to be setting the ```pulp3_content``` on ```Pulp2Content``` objects that have been migrated: https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/plugin/content.py#L290-L295

After the query

```
pulp2content.__class__.objects.bulk_update(objs=pulp2content_batch, fields=['pulp3_content'], batch_size=1000)
```
executes, I run ```pulp2content.__class__.objects.all()``` and ```pulp3_content``` is still not set on the objects, which is why I'm having this issue.

But I have no idea why this would seemingly work fine for you and for Travis, and break on my machine.  Any ideas?

pulp_2to3_migration/app/models/base.py
```diff
@@ -68,93 +72,143 @@ def get_missing_resources(self):
...
+        # flat list of all importers to migrate
+        return itertools.chain.from_iterable(
+            [plugin.repositories_importers_to_migrate for plugin in self._plugin_plans]
```
goosemania: If it happens that plan doesn't provide with any data and we chain empty list, there is still iterable and later we can't check if it's empty or not.
It creates a problem down the line when we conditionally create a mongo query.

dralley: Do you think wrapping this with a call to ```list()``` would fix that?

goosemania: yes, it should do

pulp_2to3_migration/app/migration.py
```diff
@@ -271,31 +254,34 @@ def create_repo_version(pulp3_repo_name, pulp2_repo, pulp3_remote=None):
...
-                    # re-run, a repo can be changed in pulp 2 and it might not be for the last
-                    # repo version)
-                    create_repo_version(repo_name, repo_to_migrate,
-                                        pulp2_importer_repo.pulp2importer.pulp3_remote)
+    for plugin in plan.get_plugin_plans():
+        pulp3_repo_setup = plugin.get_repo_creation_setup()
+        if not pulp3_repo_setup:
+            # create one repo version for each pulp 2 repo
+            repos_to_migrate = Pulp2Repository.objects.filter(is_migrated=False,
+                                                              type__in=plugin.type,
```
goosemania: `plugin.type` is a string, so it should be  `type=plugin.type`

PR #67, opened 2020-01-07
Specify pulp2_distributor_repository_ids instead of distributor_ids

dralley:
closes: #5837
https://pulp.plan.io/issues/5837

### DIFF
--- a/pulp_2to3_migration/app/models/base.py
+++ b/pulp_2to3_migration/app/models/base.py
@@ -178,7 +178,21 @@ class PluginMigrationPlan:
                 # created Pulp 3 remote
                 'pulp2_importer_repository_id': "foo"
                 # list of Pulp 2 repository IDs to use as the sources for created repo versions
-                'versions': []
+                'repository_versions': [
+                    # Repository Version 1 use contents of repo "idA" and distributions from the
+                    # distributors in repos "pulp2_distributor_repository_id1" and
+                    # "pulp2_distributor_repository_id2"
+                    {
+                        "pulp2_repository_id": "idA",
+                        "pulp2_distributor_repository_ids":  [
+                            "pulp2_distributor_repository_id1", "pulp2_distributor_repository_id2"
+                        ]
+                    },
+                    {
+                        "pulp2_repository_id": "idB",
+                        "pulp2_distributor_repository_ids":  ["pulp2_distributor_repository_id3"]
+                    },
+                ]
             }
         }
         """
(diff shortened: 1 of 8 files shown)


dralley:
Not done yet, just trying to get the general approach worked out.  I'm attempting to make it work without pre-processing the distributor repos.

pulp_2to3_migration/app/models/base.py
```diff
@@ -203,10 +203,15 @@ def _parse_plugin_plan(self, repository_data):
...
 
-                    distributor_ids = repository_version.get('distributor_ids', [])
-                    self.repositories_distributors_to_migrate.extend(distributor_ids)
+                    distributor_repo_ids = repository_version.get(
+                        'pulp2_distributor_repository_ids', []
+                    )
+                    self.repositories_distributors_to_migrate.extend(distributor_repo_ids)
+
+                    repository_versions.append(
+                        {'repo_id': pulp2_repository_id, 'dist_repo_ids': distributor_repo_ids}
```
dralley: Only using a dict temporarily

jlsherrill:
how does this work if a repository has multiple distributors?

ipanova:
@jlsherrill there will be created 2+ distributions ( assuming those are 2+ distributors we migrate unlike the rsync_distributor, for example)

ipanova:
> how does this work if a repository has multiple distributors?
> 
> EDIT: for example, a yum repo would have a yum_distributor, an export_distributor, and a yum_clone_distributor (custom katello one), we only care about migrating the first.

Only distributors that are specified in the migrator will be recognized and migrated example: https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/plugin/iso/migrator.py#L41

jlsherrill:
@ipanova thanks!  all looks good, as long as anything not recognized is ignored

pulp_2to3_migration/app/plugin/iso/repository.py
```diff
@@ -69,6 +69,8 @@ class IsoDistributor(Pulp2to3Distributor):
         base_config = cls.parse_base_config(pulp2distributor, pulp2_config)
         base_config['base_path'] = pulp2_config.get(
             'relative_url', pulp2distributor.pulp2_repository.pulp2_repo_id)
+        # This fails because of course there is no pulp2_repository set on lone distributors.
+        # So what should be used instead?
```
dralley: @goosemania This PR should be almost done, but I'm not sure how we want to set up the ```base_path``` in this instance.

goosemania: @dralley ,
s/`pulp2distributor.pulp2_repository.pulp2_repo_id`/`pulp2distributor.pulp2_repo_id`/

There is pulp2_repo_id stored on the distributor model. It was added with the recent PR, same issue was for the importers.

dralley: Cool, that was the last issue. It seems to work now

PR #68, opened 2020-01-08
Add another functional test

dralley:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/models/base.py
+++ b/pulp_2to3_migration/app/models/base.py
@@ -72,9 +72,10 @@ class MigrationPlan(BaseModel):
         """
         ret = {}
         if self.plan_view.missing_repositories:
-            ret['repositories'] = self.plan_view.repositories_missing_repositories
+            ret['repositories'] = self.plan_view.missing_repositories
         if self.plan_view.repositories_missing_importers:
-            ret['repositories_missing_importers'] = self.plan_view.missing_importers
+            ret['repositories_missing_importers'] = \
+                self.plan_view.repositories_missing_importers
         if self.plan_view.repositories_missing_distributors:
             ret['repositories_missing_distributors'] = \
                 self.plan_view.repositories_missing_distributors
--- a/pulp_2to3_migration/tests/functional/test_file_migration.py
+++ b/pulp_2to3_migration/tests/functional/test_file_migration.py
@@ -1,6 +1,7 @@
+import json
 import socket
-from time import sleep
 import unittest
+
 from pulpcore.client.pulpcore import (
     ApiClient as CoreApiClient,
     Configuration,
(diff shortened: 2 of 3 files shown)


pulp_2to3_migration/app/models/base.py
```diff
@@ -72,9 +72,10 @@ def get_missing_resources(self):
         """
         ret = {}
         if self.plan_view.missing_repositories:
-            ret['repositories'] = self.plan_view.repositories_missing_repositories
+            ret['repositories'] = self.plan_view.missing_repositories
```
dralley: Mistake from earlier

pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -77,25 +93,33 @@ def setUpClass(cls):
...
-            pulp2_repo_content = RepositoryContentUnit.objects.filter(repo_id=pulp2repo.repo_id)
             repo_version_content = self.file_content_api.list(repository_version=repo_version_href)
-            self.assertEqual(pulp2_repo_content.count(), repo_version_content.count)
+            self.assertEqual(PULP_2_ISO_FIXTURE_DATA[repo_id], repo_version_content.count)
+        self.assertEqual(len(repos), self.file_repo_api.list().count)
+
+    def test_1_migrate_specific_iso_repositories(self):
+        """Test that a Migration Plan with repos specified migrates only those repos."""
+        repos = list(PULP_2_ISO_FIXTURE_DATA.keys())[:2]
+        self._do_test(repos, SPECIFIC_REPOS_MIGRATION_PLAN)
```
ipanova: if we want to compare the results with the specific plan, then i think PULP_2_ISO_FIXTURE_DATA should be ordereddict otherwise the order of the ( currently 4 repos?) is not guaranteed.

pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -18,36 +19,55 @@
...
 
-    Prints the task details and a success or failure message. Exits on failure.
+PULP_2_REPOSITORIES = Pulp2Repository.objects.all()
 
-    Args:
-        tasks_api (pulpcore.client.pulpcore.TasksApi): an instance of a configured TasksApi client
-        task_href(str): The href of the task to monitor
+PULP_2_ISO_FIXTURE_DATA = {
+    # {'file': 3, ...}
+    repo.repo_id: RepositoryContentUnit.objects.filter(repo_id=repo.repo_id).count()
```
ipanova: since we also have docker repos populated, i think we should filter pulp2_repos by iso type, to exclude docker type.

pulp_2to3_migration/tests/functional/test_file_migration.py
dralley: I don't think that's the case.  It just selects the first two repo_id, and which ones they are doesn't actually matter which ones specifically because it looks up the count later using the repo_id.  So even picking them at random should be fine.

But it should always be the same ones because  as of Python 3.6 ```dict``` is ordered, and as of 3.7 it was made part of the spec (whereas before it was an implementation detail).

PR #71, opened 2020-01-10
Do not rely on repo-type from the notes 

goosemania:
I couldn't push to your branch, @ipanova

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -186,7 +186,7 @@ async def pre_migrate_lazycatalog(content_type):
             pulp2lazycatalog = []
 
 
-async def pre_migrate_all_without_content(plan):
+async def pre_migrate_all_without_content(plan, type_to_repo_ids, repo_id_to_type):
     """
     Pre-migrate repositories, relations to their contents, importers and distributors.
 
@@ -198,6 +198,8 @@ async def pre_migrate_all_without_content(plan):
 
     Args:
         plan(MigrationPlan): A Migration Plan
+        type_to_repo_ids(dict): A mapping from a pulp 2 repo type to a list of pulp 2 repo_ids
+        repo_id_to_type(dict): A mapping from a pulp 2 repo_id to pulp 2 repo type
     """
 
     _logger.debug('Pre-migrating Pulp 2 repositories')
@@ -214,7 +216,6 @@ async def pre_migrate_all_without_content(plan):
     with ProgressReport(message='Pre-migrating Pulp 2 repositories, importers, distributors',
                         code='premigrating.repositories') as pb:
         # we pre-migrate:
-        #  - empty repos (last_unit_added is not set)
         #  - repos which were updated since last migration (last_unit_added/removed >= last_updated)
         mongo_repo_q = (mongo_Q(last_unit_added__exists=False)
                         | mongo_Q(last_unit_added__gte=last_updated_naive)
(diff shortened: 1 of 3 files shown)


goosemania:
@ipanova , I haven't tested it yet :(

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -403,16 +408,29 @@
...
+    repos_to_consider = []
+
+    for repo_type in plan.get_plugins():
+        repos_to_consider += type_to_repo_ids[repo_type]
+
+        # in case only certain repositories are specified in the migration plan
+        if repos:
+            repos_to_consider = set(repos).difference(repos_to_consider)
+
+        mongo_repo_q = mongo_Q(repo_id__in=repos_to_consider)
```
goosemania: This part is basically the same as in `pre_migrate_all_without_content`, maybe worth putting in a separate function.

pulp_2to3_migration/app/tasks/migrate.py
```diff
@@ -33,6 +36,52 @@ def migrate_from_pulp2(migration_plan_pk, validate=False, dry_run=False):
         validate (bool): If True, don't migrate unless validation is successful.
         dry_run (bool): If True, nothing is migrated, only validation happens.
     """
+
+    def get_repo_types(plan):
```
goosemania: Maybe this should be done a s a part of plan processing...
Thoughts?
@ipanova , @dralley

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -224,13 +225,16 @@
...
 
+        # filter by repo type
+        repos_to_premigrate = []
+        for repo_type in plan.get_plugins():
+            repos_to_premigrate += type_to_repo_ids[repo_type]
+
         # in case only certain repositories are specified in the migration plan
         if repos:
-            mongo_repo_q &= mongo_Q(repo_id__in=repos)
+            repos_to_premigrate = set(repos).difference(repos_to_premigrate)
```
ipanova: repos_to_premigrate should be converted to set as well.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -403,16 +408,29 @@
...
+    # filter by repo type
+    mongo_repo_q = None
+    repos_to_consider = []
+
+    for repo_type in plan.get_plugins():
+        repos_to_consider += type_to_repo_ids[repo_type]
+
+        # in case only certain repositories are specified in the migration plan
+        if repos:
+            repos_to_consider = set(repos).difference(repos_to_consider)
```
ipanova: we need to convert repos_to_consider back to list otherwise L426 will fail in case of migration of multiple plugin types

pulp_2to3_migration/app/pre_migration.py
goosemania: I'm not sure I fully followed. However I think there is no pulp3 repos involved here so there should be no concern about that.

From `type_to_repo_ids[repo_type]` we get pulp 2 repo_ids and from the plan (`repos`) we get pulp 2 repo_ids.
Do you agree?

ipanova: my plan:

```
{
    "plan": {
        "plugins": [
            {
                "repositories": [
                    {
                        "name": "file",
                        "pulp2_importer_repository_id": "file",
                        "repository_versions": [
                            {
                                "pulp2_repository_id": "file"
                            }
                        ]
                    }
                ],
                "type": "iso"
            }
        ]
    },
    "pulp_created": "2020-01-13T13:44:10.661456Z",
    "pulp_href": "/pulp/api/v3/migration-plans/c9cdad50-9a24-4f32-a552-e9486fa73b0a/"
}
```
the repos_to_premigrate would be empty

ipanova: and if i have
```
{
    "plan": {
        "plugins": [
            {
                "repositories": [
                    {
                        "name": "new-name",
                        "pulp2_importer_repository_id": "file",
                        "repository_versions": [
                            {
                                "pulp2_repository_id": "file"
                            }
                        ]
                    }
                ],
                "type": "iso"
            }
        ]
    },
    "pulp_created": "2020-01-13T13:44:10.661456Z",
    "pulp_href": "/pulp/api/v3/migration-plans/c9cdad50-9a24-4f32-a552-e9486fa73b0a/"
}

repos_to_premigrate would container `new-name`, we do not have such repo-id in pulp2 to premigrate

goosemania: Thanks for pointing to the intersection! <3

pulp_2to3_migration/app/pre_migration.py
goosemania: it should be outside of the loop, thanks

PR #72, opened 2020-01-19
Fully migrate importers and distributors on every run

goosemania:
Also inspect all importers, distributor in case some of them are requested
to be migrated without a repository

closes #5980
https://pulp.plan.io/issues/5980

closes #5852
https://pulp.plan.io/issues/5852


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -447,34 +446,38 @@
...
+            for importer_type, importer_migrator in plugin.migrator.importer_migrators.items():
+                # delete all pre-migrated importers for this plugin
+                Pulp2Importer.objects.filter(pulp2_type_id=importer_type).delete()
+                # delete all migrated Remotes for this plugin
+                for remote_model in importer_migrator.pulp3_remote_models:
+                    remote_model.objects.all().delete()
+
+                # TODO: since all Remotes are removed, RemoteArtifacts need to be updated after
+                #  new Remotes are created. Maybe it's worth to keep track of Remotes and not
+                #  wipe them out every time?
```
goosemania: Still a question, how to update/add RemoteArtifacts without going through all the content all the time.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -289,19 +271,33 @@
                   'type': repo_id_to_type[record.repo_id],
                   'is_migrated': False})
 
+    if not created:
+        # if it was marked as such because it was not present in the migration plan
+        repo.not_in_pulp2 = False
```
goosemania: Maybe a better name is needed.
Currently this flag is used to signify that a repo is either not in pulp 2 or not a part of the most recent migration plan.

dralley:
The code looks good, but the 3rd smash test fails when run in isolation.  
```
pytest --verbose ./pulp_2to3_migration/tests/functional/test_file_migration.py::TestMigrationPlan::test_3_migrate_iso_repositories_with_different_importer
```
```
self = <pulp_2to3_migration.tests.functional.test_file_migration.TestMigrationPlan testMethod=test_3_migrate_iso_repositories_with_different_importer>

    def test_3_migrate_iso_repositories_with_different_importer(self):
        """Test that a Migration Plan with different importers executes correctly."""
        repos = list(PULP_2_ISO_FIXTURE_DATA.keys())
>       self._do_test(repos, DIFFERENT_IMPORTER_MIGRATION_PLAN)

pulp_2to3_migration/tests/functional/test_file_migration.py:158: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
pulp_2to3_migration/tests/functional/test_file_migration.py:133: in _do_test
    "repository id '{}'".format(repo_id))
E   AssertionError: True is not false : Missing a Pulp 3 repository for Pulp 2 repository id 'file-many'
```

It only passes when ```test_2_migrate_all_iso_repositories``` runs first.

pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -126,3 +151,8 @@ def test_2_migrate_all_iso_repositories(self):
         """Test that a Migration Plan to mirror Pulp 2 executes correctly."""
         repos = list(PULP_2_ISO_FIXTURE_DATA.keys())
         self._do_test(repos, EMPTY_ISO_MIGRATION_PLAN)
+
+    def test_3_migrate_iso_repositories_with_different_importer(self):
+        """Test that a Migration Plan with different importers executes correctly."""
+        repos = list(PULP_2_ISO_FIXTURE_DATA.keys())
```
dralley: Since the migration plan only specifies two repositories, you need to provide only those two repos here, as test 1 does.  That should fix the test failure when run individually.  However, we really need a mechanism to clear out the Pulp 3 database between test runs.

goosemania: Yes, that's the only reason there are all 4 specified at the moment.
It's still helpful to have it, having a task succeeded without failures is also a test which is better than nothing.
I tried to add checks for the importer but it would be specific to a plan and not as generic as you were able to implement the rest, so I gave up on adding it with this pr.
Until we introduce a cleanup mechanism, I'm not sure if there are options to improve the test. I can put it before "migrate-all" and then it will be two repos.

pulp_2to3_migration/app/pre_migration.py
ipanova: what about - not_in_current_migration_plan or in_current_migration, something alone the lines

goosemania: what about not_in_plan?

dkliban: not_in_plan seems the most descriptive

dralley: +1 not_in_plan

goosemania: renamed

goosemania: renamed

pulp_2to3_migration/app/migrations/0001_initial.py
```diff
@@ -12,7 +12,7 @@ class Migration(migrations.Migration):
     initial = True
 
     dependencies = [
-        ('core', '0018_auto_20191127_2350'),
+        ('core', '0019_add_signing_service_model'),
```
jlsherrill: This migration isn't in 3.0.1, only 3.1 :(  so this can't be installed with 3.0.1

PR #73, opened 2020-01-21
Problem: pulp-2to3-migration CI often fails

mikedep333:
with pkg_resources.ContextualVersionConflict between pulpcore &
setuptools.

Solution: Regenerate with latest plugin template (so that pulpcore's
branch is configurable), and install pulpcore onto the Travis host
before pulp-2to3-migration gets installed.

Fixes: #6035
https://pulp.plan.io/issues/6035

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@
 ---
 sudo: required
 # https://docs.travis-ci.com/user/trusty-ci-environment/
-dist: xenial
+dist: bionic
 language: python
 python:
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -21,7 +21,6 @@ services:
   - postgresql
   - redis-server
   - docker
-  - mongodb
```
dkliban: we need to keep this line here. it's not part of the plugin template, but we need it.

mikedep333: @dkliban I re-added that line.

PR #81, opened 2020-01-28
Fix migrated Pulp 3 repos having names from previous plans

dralley:
Remove migrated Pulp 3 repos and repo versions at the beginning of each
run.

closes: #5981
https://pulp.plan.io/issues/5981

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -455,7 +455,7 @@ async def delete_old_resources(plan):
     """
     Delete pre-migrated and migrated data for resources which are migrated fully on every run.
 
-    Delete pre-migrated importers and distributors, migrated Remotes, Publications,
+    Delete pre-migrated importers and distributors, migrated Remotes, Repositories, Publications,
     and Distributions according to a specified Migration Plan.
 
     Args:
@@ -480,3 +480,14 @@ async def delete_old_resources(plan):
                     pub_model.objects.all().delete()
                 for distribution_model in distributor_migrator.pulp3_distribution_models:
                     distribution_model.objects.all().delete()
+            for pulp2_repo in Pulp2Repository.objects.filter(type=plugin.type, is_migrated=True):
+                if pulp2_repo.pulp3_repository_version:
+                    repo_version = pulp2_repo.pulp3_repository_version
+                    pulp2_repo.pulp3_repository_version = None
+                    pulp2_repo.is_migrated = False
+                    pulp2_repo.save()
+                    repo_version.repository.delete()
+                else:
+                    _logger.warn(
+                        "Pulp2Repository says is_migrated=True but has no pulp3_repository_version"
+                    )
--- a/CHANGES/5981.bugfix
+++ b/CHANGES/5981.bugfix
@@ -1,0 +1,1 @@
+Fixed Pulp 3 repo name when migrated twice with different plans.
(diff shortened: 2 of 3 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -480,3 +480,14 @@
                     pub_model.objects.all().delete()
                 for distribution_model in distributor_migrator.pulp3_distribution_models:
                     distribution_model.objects.all().delete()
+            for pulp2_repo in Pulp2Repository.objects.filter(type=plugin.type, is_migrated=True):
+                if pulp2_repo.pulp3_repository_version:
+                    repo_version = pulp2_repo.pulp3_repository_version
+                    pulp2_repo.pulp3_repository_version = None
+                    pulp2_repo.is_migrated = False
+                    pulp2_repo.save()
+                    repo_version.repository.delete()
```
dralley: Is there a reason we weren't already doing this?  Was it especially expensive?

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -480,3 +480,14 @@
                     pub_model.objects.all().delete()
                 for distribution_model in distributor_migrator.pulp3_distribution_models:
                     distribution_model.objects.all().delete()
+            for pulp2_repo in Pulp2Repository.objects.filter(type=plugin.type, is_migrated=True):
```
dralley: Is there any reason to not simply wipe out the premigrated repos as well?  Would simplify this code.

dralley: Apart from content units, repositories are the only thing we don't currently wipe out every time.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -480,3 +480,14 @@
...
+                if pulp2_repo.pulp3_repository_version:
```
dralley: So, this wipes out all the Pulp 3 Repositories and RepositoryVersions for pre-migrated repos with ```is_migrated```.  Probably it could be done simpler if we just wipe all the Repositories of the correct type?

dralley:
The current behavior (not this branch, but master) is that if you make a new plan with less repos than a previous plan and re-migrate, the Pulp 3 repo and repo-version stick around but the distributors and importers and publications are all removed. 

```
        {
            "is_migrated": true,
            "not_in_plan": true,
            "pulp2_object_id": "5e2f74146e50c91efcf36236",
            "pulp2_repo_id": "file2",
            "pulp3_distribution_hrefs": [],
            "pulp3_publication_href": [],
            "pulp3_remote_href": null,
            "pulp3_repository_version": "/pulp/api/v3/repositories/file/file/06331068-9de6-4e32-acaa-67f9446f63fb/versions/1/",
            "pulp_created": "2020-01-29T15:30:47.579432Z",
            "pulp_href": "/pulp/api/v3/pulp2repositories/b9eb6e93-854d-41fb-9ae1-e71aa4eecd42/"
        },
```

Do we want to keep it that way?  This PR solves the problem by deleting all the repos and repo versions from Pulp 3 and setting ```is_migrated``` back to ```false```.  If we want to maintain the current behavior then this approach won't work.  But I would think that if you re-migrate that you would want to make a fresh repo version, anyways, not just ignore it because it's already migrated.

goosemania:
The intention was not to go through and not to recreate all the repo content relations on every run. That was one of concerns on the Katello side.

The result which you see is fine, imo. The key is that `not_in_plan` set to `true`. 
If you migrate this repo again and include distributors, they will be added and and content won't be readded (if there were no changes for the content).

What do you think?

dralley:
@goosemania Ok.  How should we solve https://pulp.plan.io/issues/5981#note-3 then? ```name``` is really the only mutable field, so we could special case it... or we can ignore this bug because Justin said they aren't likely to change the name of the Pulp 3 repo often.

goosemania:
Both work for me. Let's ignore for now until someone comments on the bug.

PR #85, opened 2020-01-31
Migrate global importer/distributor settings

dralley:


pulp_2to3_migration/app/models/settings.py
```diff
@@ -0,0 +1,31 @@
+from django.db import models
+
+
+class Pulp2GlobalImporterConfig(models.Model):
```
dralley: Looking into it more, I don't think this strategy will work, I think I'll need to just store the JSON blob and then have some methods attached to the plugin migrator for dealing with it, since every plugin can support different settings for importers and distributors.

dralley: Actually, there's a couple of other questions:

* Is it even a valid assumption that this is running on the same box as ```/etc/pulp/server/plugins.conf.d/```?  Or is it possibly/likely running on a different box, just w/ filesystem access to content storage and a remote connection to the database?
* Is there a point to actually storing this in the database at all?  If the migrator can process it once and then cache the global settings for the duration of the run, then it's not needed, but possibly useful for recordkeeping.

dralley: Just now noticing the ```parse_base_config``` method...

dralley: Question 1 is pretty important so I'm going to pause working on this for now

goosemania: > Looking into it more, I don't think this strategy will work, I think I'll need to just store the JSON blob and then extend IsoImporter.migrate_to_pulp3 to deal with it, since every plugin can support different settings for importers and distributors.

I agree, importers/distributors are plugin specific, JSON blob sounds good.

goosemania: > Is it even a valid assumption that this is running on the same box as /etc/pulp/server/plugins.conf.d/? Or is it possibly/likely running on a different box, just w/ filesystem access to content storage (and maybe just the content storage) and a remote connection to the database?

Both are valid assumptions. We should consider both. For the second case, maybe we can add a new requirement - access to /etc/pulp/server/plugins.conf, what do you think?
We should also ask @jlsherrill .

> Is there a point to actually storing this in the database at all? If the migrator can process it once and then cache the global settings for the duration of the run, then it's not needed, but possibly useful for recordkeeping.

If there is no need to grab any data from mongo, then I think it's fine not to store it.

jlsherrill: @goosemania yes, there is no guarantee that it will be on the box with pulp3 running.   In katello we've moved away from these global options, so we don't even really need this IMO

PR #89, opened 2020-02-05
Handling missing plugin modules

fabricio-aguiar:
https://pulp.plan.io/issues/5820
closes #5820

### DIFF
--- a/pulp_2to3_migration/app/plugin/__init__.py
+++ b/pulp_2to3_migration/app/plugin/__init__.py
@@ -1,8 +1,23 @@
+import logging
 import pkg_resources
+
+from collections import namedtuple
+from gettext import gettext as _
+
+_logger = logging.getLogger(__name__)
 
 # { plugin_name: PluginMigratorClass }
 PLUGIN_MIGRATORS = {}
+MissingMigrator = namedtuple('MissingMigrator', 'pulp3_plugin')
 
 if not PLUGIN_MIGRATORS:
     for entry_point in pkg_resources.iter_entry_points(group='migrators'):
-        PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        try:
+            PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        except ModuleNotFoundError as exc:
+            _logger.info(_(
+                'Plugin %s is not installed in pulp3 '
+                'therefore it will not be migrated from pulp2') % exc.name
+            )
+            missing = MissingMigrator(exc.msg)
+            PLUGIN_MIGRATORS[entry_point.name] = missing
--- a/CHANGES/5820.bugfix
+++ b/CHANGES/5820.bugfix
@@ -1,0 +1,1 @@
+Handling missing plugin modules


pulp_2to3_migration/app/plugin/__init__.py
```diff
@@ -1,8 +1,19 @@
...
 PLUGIN_MIGRATORS = {}
+MissingMigrator = namedtuple('MissingMigrator', 'pulp3_plugin')
 
 if not PLUGIN_MIGRATORS:
     for entry_point in pkg_resources.iter_entry_points(group='migrators'):
-        PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        try:
+            PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        except ModuleNotFoundError as exc:
+            _logger.warning(_('%s') % exc.msg)
```
dralley: I'm unsure if it warrants a warning, since not having all the plugins supported by the migrators is a fairly normal occurrence.  Maybe log as INFO?

fabricio-aguiar: I was unsure if I should go with info or warning, but as serializer [validates if plugin is installed](https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/serializers.py#L93-L96), I think we can change it to info

ipanova: since we are sending this to info shall we write a more user friendly error message? "Plugin X is not installed in pulp3 therefore it will not be migrated from pulp2" (something along the lines)

fabricio-aguiar: done:
```
pulp: pulp_2to3_migration.app.plugin:INFO: Plugin pulp_container is not installed in pulp3 therefore it will not be migrated from pulp2
```

pulp_2to3_migration/app/plugin/__init__.py
```diff
@@ -1,8 +1,22 @@
...
-        PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        try:
+            PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        except ModuleNotFoundError as exc:
+            _logger.info(_(
+                'Plugin %s is not installed in pulp3 '
+                'therefore it will not be migrated from pulp2') % exc.name
+            )
+            missing = MissingMigrator(exc.msg)
+            PLUGIN_MIGRATORS[entry_point.name] = missing
```
goosemania: @fabricio-aguiar Why do we record which migrators are missed? Where is it used?

@dralley , I think the MP parser relies on the fact that only supported plugins/migrators are present among PLUGIN_MIGRATORS, correct?

pulp_2to3_migration/app/plugin/__init__.py
```diff
@@ -1,8 +1,22 @@
+from gettext import gettext as _
+from collections import namedtuple
+import logging
 import pkg_resources
```
goosemania: Stylistic nit-pick. I think throughout our codebase we try to put module import first and import specific items after + everything in lexical order. So it would be something like:
```

￼import logging
￼import pkg_resources

￼from collections import namedtuple
from gettext import gettext as _

```

pulp_2to3_migration/app/plugin/__init__.py
```diff
@@ -1,8 +1,22 @@
...
 # { plugin_name: PluginMigratorClass }
 PLUGIN_MIGRATORS = {}
+MissingMigrator = namedtuple('MissingMigrator', 'pulp3_plugin')
 
 if not PLUGIN_MIGRATORS:
     for entry_point in pkg_resources.iter_entry_points(group='migrators'):
-        PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        try:
+            PLUGIN_MIGRATORS[entry_point.name] = entry_point.load()
+        except ModuleNotFoundError as exc:
```
goosemania: :+1:

pulp_2to3_migration/app/plugin/__init__.py
fabricio-aguiar: I recorded due to it:
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/serializers.py#L93-L96
for having a clear message, without recording it, it was falling on:
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/serializers.py#L89-L92

dralley: I believe so: https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/models/base.py#L206

goosemania: to recap the irc discussion:
it seems like https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/models/base.py#L206 won't be reached if a migrator is not there, and the presence of not installed plugins allows to fail with a specific message at the MP validation time.

PR #92, opened 2020-02-11
Clean up for RC release

goosemania:
Renamed `type` in Pulp2to3Content to `pulp2_type` for consistency.
Fixed doc strings and added comments to calify why certain relationship is needed.
Removed migration of auto_publish configuration, it's not used.

closes #6131
https://pulp.plan.io/issues/6131


goosemania:
dalley, how about re-squashing before GA? Just in case we find some other issues on the way and will need to do it again?

PR #94, opened 2020-02-11
Specifies pulpcore version explicitly in the install script

dkliban:
This is a temporary fix which will be made permanent when plugin template is updated.

[noissue]


goosemania:
Thank you, @dkliban !

goosemania:
@dkliban , I just relaized that ~=3.0, allows any Y release higher than 0, so it won't work. I think what you meant is ~=3.0.0 to allow only < 3.1

PR #99, opened 2020-02-12
Problem: plugin-template is out-of-date,

mikedep333:
and we had to put hardcoded workarounds for its lack of functionality.

Solution: Use the latest plugin-template with values specified.

re: #6149
Problem: Plugins during Travis tag/release jobs always install the latest PyPI releases of pulpcore & other plugins
https://pulp.plan.io/issues/6149

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -66,6 +66,7 @@ images:
   - pulp-2to3-migration-${TAG}:
       image_name: pulp-2to3-migration
       tag: $TAG
+      pulpcore: pulpcore~=3.0.0
       plugins:
         - ./pulp-2to3-migration
         - pulp_file~=0.1.0
--- a/template_config.yml
+++ b/template_config.yml
@@ -2,8 +2,8 @@
 # were not present before running plugin-template have been added with their default values.
 
 additional_plugins:
-- { name: pulp_file, branch: 0.1 }
-- { name: pulp_container, branch: 1.0 }
+- { name: pulp_file, branch: 0.1, pip_version_specifier: ~=0.1.0 }
+- { name: pulp_container, branch: 1.0, pip_version_specifier: ~=1.0.0 }
 black: false
 check_commit_message: true
 cherry_pick_automation: false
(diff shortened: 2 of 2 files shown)


mikedep333:
Corresponding PR: https://github.com/pulp/plugin_template/pull/183

.travis/install.sh
```diff
@@ -66,6 +66,7 @@ images:
   - pulp-2to3-migration-${TAG}:
       image_name: pulp-2to3-migration
       tag: $TAG
+      pulpcore: pulpcore~=3.0.0
       plugins:
         - ./pulp-2to3-migration
         - pulp_file~=0.1.0
```
ipanova: @mikedep333 should not we remove here the version specified for the plugins since we got it mentioned in the pip_version_specified?

mikedep333: So I thought that you managed to fix the tag/release build by making/letting the following happen:

1. pulpcore 3.1.0 gets installed from PyPI. Although the "pulpcore" variable is not passed to the container's Dockerfile.j2, [the template makes it default to "pip install pulpcore"](https://github.com/pulp/pulpcore/blob/master/containers/images/pulp/Dockerfile.j2#L59), the latest of which is 3.1.0.
2. You let ./pulp-2to3-migration (dir from travis host) get installed, which downgrades pulpcore to 3.0.0.
3. You hardcoded older release branches of pulp_file and pulp_container to be installed, which ensures that pulpcore does not get upgraded back to 3.1.0. pulpcore stays at 3.0.0. (This is now templated by additional_plugins.item.pip_version_specifier).

Based on the above understanding, I think it is best to install pulpcore 3.0.z at the beginning so it does not get downgraded.

dkliban: @ipanova: The template_config.yml values are used to generate these plugin versions when ./plugin-template --travis command is run.

ipanova: > @ipanova: The template_config.yml values are used to generate these plugin versions when ./plugin-template --travis command is run.

aha, i see, thanks @dkliban

mikedep333:
I cannot merge. A reviewer please merge.

PR #111, opened 2020-03-08
Pre-migrate Errata.

goosemania:
Extend Pulp2Content model.
Add migration of pkglists.
Make a many-to-many relationship for erratum detail model and pulp2content
Set repo for Pulp2Content at pre-migration time for Errata only.

re #6178
https://pulp.plan.io/issues/6178


pulp_2to3_migration/app/migrations/0003_modularity_yummetadatafile_erratum.py
```diff
@@ -0,0 +1,139 @@
+# Generated by Django 2.2.11 on 2020-03-23 16:33
```
ipanova: i would rename the migration to 0003_modularity_metafile_erratum_rpm ( to capture rpm models also..)
not critical, just suggestion

PR #119, opened 2020-03-18
Performance fix

dralley:
This patch reduces migration runtime for both initial and re-migrations by around 8%

[noissue]

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -30,7 +30,6 @@ from pulpcore.plugin.stages import (
 
 from pulp_2to3_migration.app.constants import NOT_USED
 from pulp_2to3_migration.app.models import (
-    Pulp2Content,
     Pulp2Importer,
     Pulp2LazyCatalog,
 )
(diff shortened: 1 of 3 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -156,11 +156,9 @@ def __init__(self, migrator):
         If a plugin needs to have more control over the order of content migration, it should
         override this method.
         """
-
-        content_types = self.migrator.content_models.keys()
-        for ctype in content_types:
+        for ctype, cmodel in self.migrator.content_models.items():
             # we need to go through all content in case any of Remotes changed
-            pulp2content_qs = Pulp2Content.objects.filter(pulp2_content_type_id=ctype)
+            pulp2content_qs = cmodel.objects.all().prefetch_related('pulp2content')
```
dralley: We're filtering the QS by type anyway, so we might as well just pass in detail querysets to begin with.

dralley:
Initial migration: 105 seconds

initial_migration

Re-migration: 87 seconds

re_migration

pulp_2to3_migration/app/plugin/content.py
goosemania: @dralley , is it ok for many-to-many relationship as well https://github.com/pulp/pulp-2to3-migration/pull/111/files#diff-7a5490ef73b4f3a627dcbd356eb955a6R108 ?

dralley:
Doing a fresh sync and publish of centos7 looks like this:

Sync time: 59 seconds
Publish time: 23 seconds

So obviously there is still room for improvement.

pulp_2to3_migration/app/plugin/content.py
ipanova: @goosemania  I have read that prefetch_related can prefetch data for many-to-one as well as many-to-many  relationships

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -201,14 +198,14 @@ def get_remote_by_importer_id(importer_id):
                 return
             return pulp2importer.pulp3_remote
 
-        for pulp2content in batch:
-            pulp_2to3_detail_content = pulp2content.detail_model
+        for pulp_2to3_detail_content in batch:
+            pulp2content = pulp_2to3_detail_content.pulp2content
```
dralley: Pretty sure this can just be turned into ```pulp2content_set```.

PR #120, opened 2020-03-18
Add a missing index + small DB optimizations

dralley:
Initial migration time: ~100 seconds -> 95 seconds
Re-migration time: ~95 seconds -> 75 seconds

[noissue]


dralley:
New graphs:

Initial migration:

initial_migration

Re-migration:

re_migration

pulp_2to3_migration/app/migrations/0006_auto_20200318_2051.py
```diff
@@ -0,0 +1,61 @@
+# Generated by Django 2.2.11 on 2020-03-18 20:51
+
+from django.db import migrations, models
```
dralley: Seems like there are a bunch of built up missing modelling changes lacking migrations.  Is this intentional?

pulp_2to3_migration/app/plugin/rpm/xml_utils.py
```diff
@@ -39,9 +39,11 @@ def _substitute_special_chars(template):
     """
     templatetag_map = dict((sym, name) for name, sym in TemplateTagNode.mapping.items())
     symbols_pattern = '(%s)' % '|'.join(templatetag_map.keys())
-    return re.sub(symbols_pattern,
-                  lambda mobj: '{%% templatetag %s %%}' % templatetag_map[mobj.group(1)],
-                  template)
+    return re.sub(
```
dralley: If you look at the graph, the templating operation is fairly expensive for the small amount of work it's doing (6%).  How difficult would it be to do the substitution in regex?

dralley:
Updated cprofile graphs (including https://github.com/pulp/pulpcore/pull/597 as well):

Initial migration:

initial_migration

Re migration:

re_migration

pulp_2to3_migration/app/plugin/rpm/xml_utils.py
goosemania: That's a good question and needs some investigation. This code is from pulp2.

IIRC,  the main problem in pulp 2 was to distinguish between the template string which needs to be substituted vs the the string which looks exactly like a template but is a part of the commet/changelog/description/etc which should not be substituted.

pulp_2to3_migration/app/migrations/0006_auto_20200318_2051.py
goosemania: I'm not sure what you mean and I don't see any migration in this PR at the moment.

goosemania:
@dralley , do we need a migration to build indexes?

pulp_2to3_migration/app/migrations/0006_auto_20200318_2051.py
dralley: I have the file locally. The changes are all in the docker plugin

dralley:
@goosemania Yes, indexes require a migration

PR #121, opened 2020-03-20
Distribution tree models for pre-migration

dkliban:
This patch contains the 2 models needed to pre-migrate Distributions from Pulp 2.


pulp_2to3_migration/app/migrations/0006_distribution_trees.py
```diff
@@ -0,0 +1,69 @@
...
+        ),
+        migrations.DeleteModel(
+            name='Pulp2Manifest',
+        ),
+        migrations.DeleteModel(
+            name='Pulp2ManifestList',
+        ),
+        migrations.DeleteModel(
+            name='Pulp2Tag',
+        ),
```
goosemania: These are docker related changes.
@dkliban , do you have this file locally /pulp_2to3_migration/app/plugin/docker/pulp_2to3_models.py ? Do you have a docker entry point in the setup.py?

@dralley, did you see the similar docker related migrations in yours?

@ipanova , ^ fyi, in case you are aware what's going on.

ipanova: :fearful: please don't remove docker!

dkliban: I did not have pulp_container installed when I generated the migration originally. Just pushed a new version.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -415,3 +415,53 @@ def _get_profiles(profiles):
...
+    """
+    Pulp 2to3 detail content model to store Pulp2 Distribution content details.
+    """
+
+    # Unit key fields
+    distribution_id = models.TextField()
+    family = models.TextField()
+    variant = models.TextField()
+    version = models.TextField()
+    arch = models.TextField()
```
goosemania: I wonder if that's enough.
How will you identify the location of all files which belong to the distribution tree.
I *think* you can get this info from `files` field in pulp2 but I'm not sure.

PR #122, opened 2020-03-24
Performance cherry-picks for 0.1

goosemania:


dralley:
@goosemania Keep in mind that that last patch requires the user to rerun the migrations

PR #136, opened 2020-04-03
Adds migration of RPM Distributions to DistributionTrees

dkliban:
This patch comes with a new migration that adds pulp3_repository field to the Pulp2Repository
model. This is needed for creating a Variant associated with a DistributionTree in Pulp 3.

The Addons are completely ignored because pulp 2 did not support them.

Variants are only limited to the ones that reference the top level repository ('.').

closes: #6201
https://pulp.plan.io/issues/6201


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -229,13 +231,49 @@ def get_remote_by_importer_id(importer_id):
...
+                    artifact=Artifact(),
+                    url=urljoin(remote.url, relative_path),
+                    relative_path=relative_path,
+                    remote=self.remote,
+                    deferred_download=False,
+                )
+                d_artifacts.append(da)
+                dc.extra_data = future_relations
+                dc = DeclarativeContent(content=pulp3content, d_artifacts=d_artifacts)
+                await self.put(dc)
```
goosemania, line 269: I think this all should be done per each importer_id from lazycatalog.
Imagine you sync the same content through different importers in pulp2.
Right now you are creating artifacts for all the importers and then a treeinfo file for one of them.

Maybe stages API will handle the same artifatcs with different Remotes correctly and the same artifacts will be deduplicated on the way. In this case we need treeinfo files with all the remotes created as well.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -229,13 +231,49 @@ def get_remote_by_importer_id(importer_id):
...
+                    d_artifacts.append(da)
+
+                # We do this last because we need the remote url which is only found in the LCE
+                # of the image files. There is no LCE for the .treeninfo file itself.
+                relative_path = pulp_2to3_detail_content.relative_path_for_content_artifact
+                da = DeclarativeArtifact(
+                    artifact=Artifact(),
+                    url=urljoin(remote.url, relative_path),
+                    relative_path=relative_path,
+                    remote=self.remote,
```
goosemania, line 263: The case when there is no remote available is not handled.
E.g. immediate policy case, or any other reason for not having lazy catalog entries.

pulp_2to3_migration/app/plugin/content.py
dkliban: I've added tracking of Remotes.

pulp_2to3_migration/app/plugin/content.py
dkliban: I've added a case for non on_demand content.

pulp_2to3_migration/app/plugin/content.py
dralley: Should you tag this issue then? https://pulp.plan.io/issues/6375

dkliban: @dralley My changes don't address that issue.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -6,10 +6,12 @@
 from django.contrib.postgres.fields import JSONField
 from django.db import models
 
-from pulp_2to3_migration.app.models import Pulp2to3Content
+from pulp_2to3_migration.app.models import Pulp2to3Content, Pulp2Repository
```
goosemania, line 9: nitpick, when it's more than 1 import, we tend to put them on the multiple lines
```
from X import (
  A,
  B,
)
```

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -21,6 +23,8 @@
     UpdateRecord,
 )
 
+from pulp_rpm.app.kickstart.treeinfo import PulpTreeInfo, TreeinfoData
```
goosemania, line 26: Same here

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -535,7 +546,28 @@ async def create_pulp3_content(self):
...
+            treeinfo = PulpTreeInfo()
+            try:
+                treeinfo.load(f=os.path.join(self.pulp2content.pulp2_storage_path, '.treeinfo'))
+            except FileNotFoundError:
+                continue
+            self.filename = namespace
+            treeinfo_parsed = treeinfo.parsed_sections()
+            treeinfo_serialized = TreeinfoData(treeinfo_parsed).to_dict(filename=namespace)
+            # The repository must already be migrated in order to migrate the DistributionTree
+            pulp2_r_c_unit = RepositoryContentUnit.objects.get(unit_id=self.pulp2content.pulp2_id)
```
goosemania, line 560: As a rule of thumb, we are trying not to query MongoDB after pre-migration. You should have all the data needed available in PostgreSQL at this point.
The data you need is pre-migrated into Pulp2RepoContent table/model.

pulp_2to3_migration/app/models/repository.py
```diff
@@ -56,6 +57,11 @@ class Pulp2Repository(BaseModel):
                                                 on_delete=models.SET_NULL,
                                                 null=True)
 
+    # This is needed for migrating Variants of the DistributionTree
+    pulp3_repository = models.ForeignKey(Repository,
```
goosemania, line 61: Do we need a custom migration for users to upgrade from older versions of the migration plugin? (e.g. those who migrated Docker and File already, say with Katello 3.15.)
I guess it's not essential if we are not exposing this field in REST API.
It's just somewhat inconsistent that some migrated repos will have this field set and some won't.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -535,7 +546,28 @@ async def create_pulp3_content(self):
...
+            pulp2_repo_id = pulp2_r_c_unit.repo_id
+            pulp2_repo = Pulp2Repository.objects.get(pulp2_repo_id=pulp2_repo_id)
```
goosemania, line 562: Since pulp2_repo_id is not unique in the span of multiple migration re-runs (a repo in pulp2 can be removed and a new, completely different one, can be created with the same repo_id), an additionla condition is needed: `not_in_plan=False`.
It means that you are looking for a repository which is expected to be present/migrated as a part of the current migration plan.

goosemania:
I tested with  http://mirror.linux.duke.edu/pub/centos/8/BaseOS/x86_64/kickstart/ and it fails like this:
```
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: future: <Task finished coro=<RpmMigrator.migrate_content_to_pulp3() done, defined at /home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/rpm/migrator.py:135> exception=KeyError('../../../AppStream/x86_64/os/',)>
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: Traceback (most recent call last):
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/rpm/migrator.py", line 142, in migrate_content_to_pulp3
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: await dm.create()
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/content.py", line 83, in create
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: await pipeline
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py", line 228, in create_pipeline
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: await asyncio.gather(*futures)
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py", line 43, in __call__
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: await self.run()
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/content_stages.py", line 121, in run
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: await self._post_save(batch)
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulp_rpm/pulp_rpm/app/tasks/synchronizing.py", line 960, in _post_save
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: _handle_distribution_tree(declarative_content)
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: File "/home/vagrant/devel/pulp_rpm/pulp_rpm/app/tasks/synchronizing.py", line 912, in _handle_distribution_tree
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: resource["repository_id"] = treeinfo_data["repositories"][key]
Apr 15 17:55:29 pulp2-nightly-pulp3-source-centos7.rhgoose.example.com rq[31449]: KeyError: '../../../AppStream/x86_64/os/'

```

Even though the error is in the post_save in the RPM plugin, the daily performance tests builds a re not failing and they are syncing/publishing this repo as well. So I suspect that it might be a migration problem.

pulp_2to3_migration/app/models/repository.py
dkliban: This field is only used by the RPM plugin. So I am not sure we need to do anything custom here.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -229,13 +231,75 @@ def get_remote_by_importer_id(importer_id):
...
+                dc.extra_data = future_relations
+                await self.put(dc)
             # not all content units have files, create DC without artifact
             if is_artifactless_type:
                 # dc without artifact
                 dc = DeclarativeContent(content=pulp3content)
                 dc.extra_data = future_relations
                 await self.put(dc)
-            else:
+            elif not is_multi_artifact:
```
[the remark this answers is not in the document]
dkliban: That makes sense. I'll update it.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -229,13 +231,75 @@ def get_remote_by_importer_id(importer_id):
...
                 # create artifact for content that has file
                 # TODO potentially improve this part for content that can have multiple files
```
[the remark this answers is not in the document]
dkliban: Good call. I'll remove it.

pulp_2to3_migration/app/plugin/rpm/migrator.py
```diff
@@ -88,6 +91,7 @@ class RpmMigrator(Pulp2to3PluginMigrator):
     pulp3_plugin = 'pulp_rpm'
     pulp3_repository = pulp3_models.RpmRepository
     content_models = OrderedDict([
+        ('distribution', Pulp2Distribution),
```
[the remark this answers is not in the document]
dkliban: I was just going with the alphabetical order here. I'll keep the RPMs first.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
dkliban: Good point. That actually makes this even simpler. Thanks!

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
dkliban: Cool. I'll add that condition.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -570,6 +570,12 @@ async def create_pulp3_content(self):
             treeinfo_serialized["repositories"] = {'.': pulp2_repo.pulp3_repository.pk}
             # Pulp 2 did not support addon repositories, so we should not list them here either
             treeinfo_serialized['addons'] = []
+            # Pulp 2 only supported variants that are in the root of the repository
+            variants = []
+            for variant in treeinfo_serialized['variants']:
+                if variant['repository'] == '.':
```
goosemania, line 576: Can there be more than one? I think it's a path to an RPM repo, so there can only be one in '.'. What do you think?

dkliban: I was not sure if there can be multiple variants that point to the same repo.

PR #140, opened 2020-04-21
Bump version for pulpcore dependency

goosemania:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,12 +3,10 @@
 from setuptools import find_packages, setup
 
 requirements = [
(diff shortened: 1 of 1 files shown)


goosemania:
I think we can't require createrepo_c because not every installation will have rpm plugin.
The only thing we can do is indirectly require certain plugin versions through a pulpcore requirement :/
Aka we are requiring pulpcore 3.3, it means we need pulp_rpm compatible with pulpcore 3.3 (and this rpm version has the createrepo_c needed)

PR #142, opened 2020-04-21
Parallel tasks

dralley:
Adds a parallel strategy for migrating repositories without making additions to progress reporting (but does remove some progress reporting because it was difficult to maintain).

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -5,7 +5,7 @@ from pulpcore.plugin.models import (
     CreatedResource,
     ProgressReport,
     Repository,
-    TaskGroup,
+    Task,
 )
 from pulpcore.plugin.tasking import enqueue_with_reservation
 
@@ -253,9 +253,6 @@ def create_repoversions_publications_distributions(plan, parallel=True):
             task_func = complex_repo_migration
 
             if parallel:
-                task_group = TaskGroup.objects.create(
-                    description="Migration Sub-tasks"
-                )
                 for repo_name in pulp3_repo_setup:
                     repo = Repository.objects.get(name=repo_name).cast()
                     task_args = [plugin, pulp3_repo_setup, repo_name]
@@ -263,10 +260,8 @@ def create_repoversions_publications_distributions(plan, parallel=True):
                         task_func,
                         [repo],
                         args=task_args,
-                        task_group=task_group
+                        task_group=Task.current().task_group
                     )
-                resource = CreatedResource(content_object=task_group)
-                resource.save()
             else:
                 # Serial (non-parallel)
                 for repo_name in pulp3_repo_setup:
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -4,7 +4,10 @@ from pulpcore.plugin.models import (
     Content,
     CreatedResource,
     ProgressReport,
+    Repository,
+    TaskGroup,
 )
+from pulpcore.plugin.tasking import enqueue_with_reservation
 
 from pulp_2to3_migration.app.models import (
     Pulp2Content,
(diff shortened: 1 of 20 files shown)


pulp_2to3_migration/app/migration.py
```diff
@@ -106,216 +143,194 @@ async def migrate_repositories(plan):
...
 
-        for pulp2importer in pulp2importers_qs:
-            importer_migrator = importer_migrators.get(pulp2importer.pulp2_type_id)
-            remote, created = await importer_migrator.migrate_to_pulp3(pulp2importer)
-            pulp2importer.pulp3_remote = remote
-            pulp2importer.is_migrated = True
-            pulp2importer.save()
-            if created:
-                pb.increment()
+            if parallel:
```
dralley, line 244: The rest is just refactoring, but the meat of the PR is basically just these 2 dozen lines.  This threadpool code is the experiment I was talking about that only took about an hour to research and implement.

I did some not very scientific benchmarking to get an idea whether it might be nearly as fast or faster, here are the results:

Thread Pool

    Intial migration: 243 seconds
    Re-migration: 94 seconds

Multiple Tasks

    Intial migration: 245 seconds
    Re-migration: 94 seconds

Serialized (no parallel)

    Intial migration: 258 seconds
    Re-migration: 101 seconds

dralley: Don't read too far into those results though, like I said, it was not very scientific.  I used Centos 6, 7, and 8-baseos, but I also had Fedora 30, 31, and Centos 8-appstream synced into Pulp 2 so these numbers are migrating all existing content from all 6 repos + those 3 repos I mentioned.  I gave the VM 5 CPUs, 12gb of memory, and ran 4 Pulp workers.  I also ran ```htop``` and I noticed that even with 12gb of memory and only 3 repos, it was still using swap space, which might mess with the numbers.

dralley:
I have another PR (unpushed) that removes async entirely except for where it is being used to migrate content.  It knocks ~5 seconds off of both migration and re-migration, which is not really a big deal, but if we aren't gaining anything from it then we may as well get rid of it.

pulp_2to3_migration/app/migration.py
```diff
@@ -236,86 +241,110 @@ async def create_repo_versions(plan):
...
+                #     for repo_name in pulp3_repo_setup:
+                #         task_args = [plugin, pulp3_repo_setup, repo_name]
+                #         executor.submit(task_func, *task_args)
+
+                # Multi-task strategy
+                for repo_name in pulp3_repo_setup:
+                    task_args = [plugin, pulp3_repo_setup, repo_name]
+                    enqueue_with_reservation(
+                        task_func,
+                        [],  # should we be locking on the repo?
```
goosemania, line 271: +1 to lock on a repo

dralley:
@goosemania @ipanova @dkliban In keeping with our discussion yesterday I'm going to enable the threadpool, leave the code for multiprocess present (but commented out), and then we can evaluate progress reporting once the import/export team is in a position to think about their needs in this area.

dralley:
Still one minor issue that I caught manually (tests don't catch much).

dralley:
@dkliban @goosemania @ipanova 

I think this is most of the migration plugin work, if not all of it.  The task group gathering stuff needs to happen in Pulpcore. This is ready for review, let me know if I'm missing anything.

We will also need the following two Pulpcore PRs merged before this is fully functional

* https://github.com/pulp/pulpcore/pull/677
* https://github.com/pulp/pulpcore/pull/672

dralley:
I went ahead and removed async from everywhere except for the content migration stage where it is actually being used.  Basically needing to run things on the event loop gets in the way of parallelizing, because you can't send async functions directly to the tasking system, you have to have an outer synchronous function and create new event loops inside. It makes life needlessly more difficult.

 Since we can't benefit from it anyway, and since it's possibly dangerous due to unknown ORM-async interactions, and since it adds a small amount of extra memory + CPU overhead to perform the scheduling of the tasks that execute in linear order anyways, it's better to just get rid of it entirely.

I tested the performance again afterwards, it is exactly as fast as it was before, or maybe a fraction of a percent faster.

goosemania:
I'm concerned that some changes needed in the pulpcore. We need to be compatible with 3.3.
Katello is on 3.3. If changes can go into 3.3.z, I believe it's fine. If something will go into 3.4, it's a problem (to my knowledge).

dralley:
@goosemania Both of those PRs can go into 3.3.z, the latter is a bugfix and the former is fixing something that should have been in 3.3.0 but mistakenly wasn't.

As far as progress reporting goes, they are small enough and well-contained enough changes that I wouldn't feel too bad about slipping them into a Z-release.

PR #148, opened 2020-05-12
Update travis

goosemania:
[noissue]


dkliban:
The hostname is now 'pulp' and the API is available on port 80. So update this line here: https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/tests/functional/test_file_migration.py#L121 to http://pulp

PR #149, opened 2020-05-14
Add pulp_deb plugin support (Packages only for now)

quba42:
closes #7863
https://pulp.plan.io/issues/7863


pulp_2to3_migration/app/migrations/0012_auto_20200513_1311.py
```diff
@@ -0,0 +1,170 @@
+# Generated by Django 2.2.12 on 2020-05-13 13:11
```
quba42, line 1: Why is `django-admin makemigrations` adding this file (even on `master` before I changed anything)?
Is it because I do not have the `pulp_rpm` plugin installed in my box?
Weird.

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -0,0 +1,50 @@
...
+        'deb_distributor': DebDistributor,
+    }
+
+    @classmethod
+    def migrate_content_to_pulp3(cls):
+        """
+        Migrate pre-migrated Pulp 2 Debian content.
+        """
+        first_stage = ContentMigrationFirstStage(cls)
+        dm = DeclarativeContentMigration(first_stage=first_stage)
```
quba42, line 46: So so far, I am simply trying to use the default migration pipeline. (The same one that `pulp_file`/`iso` uses.)
I am not sure if this naive approach could possibly work, since I am not that well versed in the topic of pulp3 async pipelines.
Any pointers are appreciated.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -0,0 +1,165 @@
...
+    content creation.
+    """
+    pulp2_type = 'deb'
+    name = django_models.CharField(max_length=255)
+    version = django_models.CharField(max_length=255)
+    architecture = django_models.CharField(max_length=255)
+    checksumtype = django_models.CharField(max_length=255)
+    checksum = django_models.CharField(max_length=255)
+    filename = django_models.CharField(max_length=255)
+    size = django_models.PositiveIntegerField()
```
quba42, line 22: This Integer field (and/or the `installed_size` one below), appear to be causing me grief.
I keep getting `TypeError: int() argument must be a string, a bytes-like object or a number, not 'DeferredAttribute'` on the first package the pipeline tries to migrate.
When I change these fields to `django_models.TextField` (and then manually cast things back to `int` in the code), the pipeline gets a bit further (I can see some looping over all the packages I am trying to migrate) but ultimately fails with the same error.

I am somewhat at my wits end about this. What I am doing here looks similar to what rpm is doing to me...
Maybe I am using the async pipeline wrong, or maybe it is something small and trivial I am overlooking.

pulp_2to3_migration/app/migrations/0012_auto_20200513_1311.py
goosemania: Yes, I believe it's because you don't have pulp_rpm.
If pulp_rpm is not installed, entry_point for it is ignored, thus django can't read models of that plugin and tries to remove them from the db.

pulp_2to3_migration/app/plugin/deb/migrator.py
goosemania: It can work, it depends on your content.
How simple is your sync workflow: do you have a custom sync pipeline? Does order of content imported matters? Do you need to build any relations between different content units after they are in Pulp?

goosemania: I looked closer at the code you are adding. Since for now it's just a debian package migration. I think it should work with default first stage and pipeline.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
goosemania: I agree it's strange. Hard to suggest something without any traceback or an attempt to reproduce.
@dralley , @dkliban , any thoughts?

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -0,0 +1,165 @@
...
+            'package': self.name,
+            'source': self.source,
+            'version': self.version,
+            'architecture': self.architecture,
+            'section': self.section,
+            'priority': self.priority,
+            'installed_size': self.installed_size,
+            'maintainer': self.maintainer,
+            'original_maintainer': self.original_maintainer,
+            'description': self.description,
```
dralley, line 149: @quba42 I notice that the "size" field is not present here, and I think it probably should be?

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
dralley: @quba I left a comment down below, I don't know if it is the cause of your problem, but it seems at least plausible that it might be

quba: Wrong quba :) ping @quba42

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -0,0 +1,50 @@
...
+
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(dm.create())
+        loop.close()
```
dralley, line 50: Delete this line in accordance with https://github.com/pulp/pulp-2to3-migration/pull/151/files

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
dralley: It's plausible that creating a Package object without initializing the ```size``` field might cause problems when the uninitialized field is saved.  I would have expected a more descriptive error if that were the case, but maybe not.

quba42: This is a good hint. I think I need to look more closely at how the `Content` and the `Artifact` classes interact. I am not sure the size field needs to be added here directly (does not appear to work that way for iso and rpm) but I may be missing some mechanism for it to be passed to the artifact that is created.

I am on holiday this week, but I will definitely pursue this when I get back, thanks.

pulp_2to3_migration/app/plugin/deb/migrator.py
quba42: Will do, I generally try to check all incoming commits whenever I rebase to a newer `master` :wink:.

pulp_2to3_migration/app/plugin/deb/migrator.py
quba42: My plan/hope is that I can get only packages to work in a first step (keeping it as simple as possible), and then look into "Releases" and "Components" in a second step (at which point I will most likely need to adjust the pipeline).
Repositories containing only packages have an independent use case in pulp_deb so I think this approach makes sense.

pulp_2to3_migration/app/migrations/0012_auto_20200513_1311.py
quba42: I will just leave this thread as a reminder to clean up/regenerate the migrations at some point before I convert this to "ready for review".

pulp_2to3_migration/app/plugin/deb/migrator.py
quba42: Should be done as part of a rebase to current master.

pulp_2to3_migration/app/migrations/0012_auto_20200513_1311.py
quba42: Finally built a pulplift box with all relevant plugins and regenerated the migrations.
So I can resolve this thread.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
quba42: So this error appears to have gone away for now, so I am closing this thread for now.

quba42:
The latest error I am stuck on is:
```
pulp: pulp_2to3_migration.pulp2.connection:INFO: Attempting to connect to localhost:27017
pulp: rq.worker:ERROR: ValueError: No declared artifact with relative path "tuxedo-cc-wmi_0.1.4_all.deb" for content "<Package: __>"
Traceback (most recent call last):
File "/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/worker.py", line 883, in perform_job
File "/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py", line 657, in perform
File "/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py", line 663, in _execute
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/tasks/migrate.py", line 140, in migrate_from_pulp2
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/migration.py", line 36, in migrate_content
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/plugin/deb/migrator.py", line 49, in migrate_content_to_pulp3
File "/usr/lib64/python3.6/asyncio/base_events.py", line 484, in run_until_complete
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/plugin/content.py", line 84, in create
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py", line 225, in create_pipeline
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py", line 43, in __call__
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/artifact_stages.py", line 225, in run
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/artifact_stages.py", line 269, in _needed_remote_artifacts
ValueError: No declared artifact with relative path "tuxedo-cc-wmi_0.1.4_all.deb" for content "<Package: __>"
Traceback (most recent call last):
File "/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/worker.py", line 883, in perform_job
File "/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py", line 657, in perform
File "/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py", line 663, in _execute
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/tasks/migrate.py", line 140, in migrate_from_pulp2
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/migration.py", line 36, in migrate_content
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/plugin/deb/migrator.py", line 49, in migrate_content_to_pulp3
File "/usr/lib64/python3.6/asyncio/base_events.py", line 484, in run_until_complete
File "/home/vagrant/devel/pulp_2to3_migration/pulp_2to3_migration/app/plugin/content.py", line 84, in create
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py", line 225, in create_pipeline
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py", line 43, in __call__
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/artifact_stages.py", line 225, in run
File "/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/artifact_stages.py", line 269, in _needed_remote_artifacts
ValueError: No declared artifact with relative path "tuxedo-cc-wmi_0.1.4_all.deb" for content "<Package: __>"
```
At this point I am thinking that my problem is that I just don't understand the central async pipelines that are being used...

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -0,0 +1,157 @@
...
+                    pulp2content=pulp2_id_obj_map[deb['_id']],
+                    size=deb['size'],
+                    **pre_migrate_fields,
+                )
+            )
+
+        cls.objects.bulk_create(pulp2deb_to_save, ignore_conflicts=True, batch_size=DEFAULT_BATCH_SIZE)
+
+    @classmethod
+    def create_pulp3_content(self):
```
quba42, line 132: As far as I can tell everything works fine until `create_pulp3_content` is called.
However within this function I can't seem to access any of the fields of the pre migrated units being passed to it.
Every field access (e.g.: `self.checksum`) appears to return a `DeferredAttribute` object, rather than a string value.

quba42: I added a `self.checksum` call and printed the result to log output, and got:
```
<django.db.models.query_utils.DeferredAttribute object at 0x7f056bb8a8d0>
```

quba42: Within the DB things look right to me, e.g.:
```
pulp=# SELECT checksum FROM pulp_2to3_migration_pulp2debpackage WHERE filename='nvidia-container-toolkit_1.0.5-1_amd64.deb';
-[ RECORD 1 ]--------------------------------------------------------------
checksum | 9ffe51f7520528ef9215a556abb641a16dbbc1e8d44b8ea2c413e0f711de21ff
```

dralley: @quba42 The bug is actually right on this very line :)

You've defined this method as a classmethod, which passes the class through the ```self``` parameter.  Therefore when you access attributes on ```self```, they exist, but the value you get back is the type they exist as on the class, as opposed to the type you'd get when accessing an instance.

It should just be a normal method, not a classmethod.  It might have used to have been a classmethod though, I can't remember.

quba42: That has done it! I was now able to successfully migrate Debian packages from Pulp2 to Pulp3!
There is certainly some more work to be done, but the rest of it should be a lot more straight forward.
This very much concludes an important initial milestone!

quba42:
I believe this is now ready for an initial review.
It has some limited but independently useful functionality.
The current state is able to migrate Debian packages from Pulp2 to Pulp3.
The so created repository is published using the simple publisher to create a usable Pulp3 publication. (I have not yet been able to look over the resulting publication due to some network configuration issue with my development environment...)

What is currently missing is all the content types required for structured publishing.
It might make sense to review and merge the current state, and then extend its functionality on separate issue branches (to avoid a single giant PR seeking to add everything at once).
However, in that case it needs to be made clear that the current state is not full featured or ready for production.
How was this handled for other plugins?

quba42:
I would of course be happy to arrange a meeting/screenshare or similar on how I am currently testing this in my development environment.

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -0,0 +1,184 @@
...
+            QueryExistingContents(),
+            ContentSaver(),
+            ResolveContentFutures(),
+            RelatePulp2to3Content(),
+        ]
+
+        return pipeline
+
+
+class DebContentMigrationFirstStage(ContentMigrationFirstStage):
```
goosemania, line 103: @quba42 , could you share what pulp_deb needs here that the default implementation doesn't cover? Why do you need  to override `migrate_to_pulp3`?

quba42: Yes, actually I had meant to comment on that. I separated that out in order to better understand it (strip out the many if clauses that I was not actually using. However, at this point I don't think I am actually using anything I could not achieve with `ContentMigrationFirstStage`. I will retest to make sure, and remove this if it works without.

I may need to add it back in when I add the additional content types.

quba42: It works without, so I dropped this subclassing. I also fixed a couple of linter errors.

quba42:
Rebased to take advantage of Travis pipeline with `pulp_deb` installed in it.

pulp_2to3_migration/app/plugin/deb/pulp2_models.py
```diff
@@ -0,0 +1,174 @@
+import mongoengine
+
+from pulp_2to3_migration.pulp2.base import (
+    FileContentUnit,
+    ContentUnit,
```
goosemania, line 5: Same here ;)

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -0,0 +1,92 @@
...
+)
+
+from pulp_deb.app import models as pulp3_models
+
+from pulp_2to3_migration.app.plugin.api import (
+    Pulp2to3PluginMigrator,
+    ContentMigrationFirstStage,
+    RelatePulp2to3Content,
+    DeclarativeContentMigration,
+)
```
goosemania, line 18: We try to keep imports according to pep8 and in alphabetical order, both modules imported from and the object which are being imported.
```
from pulpcore.plugin.stages import (
￼    ArtifactSaver,
￼    ContentSaver,
    ...
)
from pulp_2to3_migration.app.plugin.api ...
from pulp_deb.app ...
```
It would be nice if you could check imports in other place which you are adding as well.

pulp_2to3_migration/app/plugin/deb/pulp2_models.py
```diff
@@ -0,0 +1,174 @@
...
+        multi_arch="Multi-Arch",
+        homepage="Homepage",
+        description="Description",
+        original_maintainer="Original-Maintainer",
+    )
+
+
+class DebComponent(ContentUnit):
+    """
+    This unittype represents a deb release/distribution component
```
goosemania, line 122: Nit-pick. For the sake of consistency, it would be good to add period at the end of the doc strings. Here and in other places as well.

goosemania (changes requested):
@quba42, I have just some minor comments, other than that it's ready to merge I believe. Thanks  for working on it!

goosemania:
@quba42, and about the plugin support, I suggest to add a line to a README and to the docs. 
Either that pulp_deb is in active development and currently supporting packages only, or mention that it's a tech preview with packages suppor tonly.

pulp_2to3_migration/app/plugin/deb/migrator.py
quba42: Done.

pulp_2to3_migration/app/plugin/deb/pulp2_models.py
quba42: Done.

quba42:
I squished all commits into one and added an issue reference in the commit message.
I also added a line on plugin support to both the README and the main documentation.
In addition I added some more detailed description of what does and does not work to the workflow documentation.

quba42:
Not sure why the tests are now failing, looks a bit like a timeout issue to me.

CHANGES/7863.feature
```diff
@@ -0,0 +1 @@
+Added support to migrate Debian packages (tech preview).
```
goosemania, line 1: :+1:

goosemania (approved):
Great job!

goosemania:
> Not sure why the tests are now failing, looks a bit like a timeout issue to me.

I restarted the job, we'll see.

PR #151, opened 2020-05-19
Fix crashing issue w/ multiple plugin type migrations

dralley:
closes: #6754
https://pulp.plan.io/issues/6754


dralley:
We can insert something to clean up the event loop once the content stages are done with it, it'll just look a little out of place since it was originally created in an entirely different file.  I'm not immediately super concerned about it. @bmbouter what do you think?

bmbouter:
Letting the loop close automatically seems ok to me.

PR #152, opened 2020-05-21
Track distributors, recreate Publications/Distributions only if needed

goosemania:
closes #6376
https://pulp.plan.io/issues/6376

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -52,6 +52,8 @@ def migrate_repositories(plan):
     )
     with ProgressReport(**progress_data) as pb:
         for plugin in plan.get_plugin_plans():
+
+            # TODO: Should we filter out repos which have is_migrated=True?
             pulp2repos_qs = Pulp2Repository.objects.filter(
                 pulp3_repository_version=None,
                 not_in_plan=False,
@@ -191,6 +193,7 @@ def complex_repo_migration(plugin, pulp3_repo_setup, repo_name):
             pass
     for pulp2_repo_info in repo_versions_setup:
         try:
+            # TODO: should we check the is_migrated=True?
             pulp2_repo = Pulp2Repository.objects.get(
                 pulp2_repo_id=pulp2_repo_info['repo_id'],
                 not_in_plan=False
@@ -217,8 +220,9 @@ def complex_repo_migration(plugin, pulp3_repo_setup, repo_name):
             continue
         else:
             pulp2dist = Pulp2Distributor.objects.filter(
+                is_migrated=False,
                 pulp2_repo_id__in=dist_repositories,
-                pulp2_type_id__in=distributor_types
+                pulp2_type_id__in=distributor_types,
             )
             for dist in pulp2dist:
                 dist_migrator = distributor_migrators.get(dist.pulp2_type_id)
@@ -281,6 +285,7 @@ def create_repo_version(migrator, pulp3_repo_name, pulp2_repo, pulp3_remote=None
         pulp3_remote(remote): a pulp3 remote
     """
 
+    # TODO: adjust either the comment or the code due to tracking of remotes
     # Add a remote to every repo, even the migrated one, because remotes are migrated on
     # every run
     if pulp3_remote:
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -193,7 +193,6 @@ def complex_repo_migration(plugin, pulp3_repo_setup, repo_name):
             pass
     for pulp2_repo_info in repo_versions_setup:
         try:
-            # TODO: should we check the is_migrated=True?
             pulp2_repo = Pulp2Repository.objects.get(
                 pulp2_repo_id=pulp2_repo_info['repo_id'],
                 not_in_plan=False
@@ -214,7 +213,8 @@ def complex_repo_migration(plugin, pulp3_repo_setup, repo_name):
 
         try:
             migrated_repo = Pulp2Repository.objects.get(pulp2_repo_id=repo_id,
-                                                        not_in_plan=False)
+                                                        not_in_plan=False,
+                                                        is_migrated=True)
         except Pulp2Repository.DoesNotExist:
             # not in Pulp 2 anymore
             continue
@@ -276,7 +276,8 @@ def create_repo_version(migrator, pulp3_repo_name, pulp2_repo, pulp3_remote=None
     """
     Create a repo version based on a pulp2 repository.
 
-    Add a remote to a corresponding pulp 2 repository.
+    Add a remote to a corresponding pulp 2 repository. Since any remote can change without a repo
+    being changed itself, re-set it here for every repo.
 
     Args:
         migrator: migrator to use, provides repo type information
@@ -285,9 +286,6 @@ def create_repo_version(migrator, pulp3_repo_name, pulp2_repo, pulp3_remote=None
         pulp3_remote(remote): a pulp3 remote
     """
 
-    # TODO: adjust either the comment or the code due to tracking of remotes
-    # Add a remote to every repo, even the migrated one, because remotes are migrated on
-    # every run
     if pulp3_remote:
         pulp2_repo.pulp3_repository_remote = pulp3_remote
     # pulp2importer might not be migrated, e.g. config was empty
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -232,7 +232,8 @@ class ContentMigrationFirstStage(Stage):
             lazy_wo_catalog = is_lazy_type and not pulp2lazycatalog
             if pulp2content.pulp3_content is not None and (not is_lazy_type or lazy_wo_catalog):
                 if pb:
-                    pb.increment()
+                    pb.total -= 1
+                    pb.save()
                 continue
             if is_lazy_type and not pulp2content.downloaded and not pulp2lazycatalog:
                 _logger.warn(_('On_demand content cannot be migrated without an entry in the '
--- a/pulp_2to3_migration/app/plugin/docker/repository.py
+++ b/pulp_2to3_migration/app/plugin/docker/repository.py
@@ -55,7 +55,10 @@ class DockerDistributor(Pulp2to3Distributor):
         base_config['base_path'] = pulp2_config.get(
             'repo-registry-id', pulp2distributor.pulp2_repo_id)
         base_config['repository_version'] = repo_version
-        distribution, created = ContainerDistribution.objects.update_or_create(**base_config)
+        distribution, created = ContainerDistribution.objects.update_or_create(
+            name=base_config['name'],
+            base_path=base_config['base_path'],
+            defaults=base_config)
         return None, distribution, created
 
     @classmethod
--- a/pulp_2to3_migration/app/plugin/docker/repository.py
+++ b/pulp_2to3_migration/app/plugin/docker/repository.py
@@ -57,3 +57,41 @@ class DockerDistributor(Pulp2to3Distributor):
         base_config['repository_version'] = repo_version
         distribution, created = ContainerDistribution.objects.update_or_create(**base_config)
         return None, distribution, created
+
+    @classmethod
+    def needs_new_publication(cls, pulp2distributor):
+        """
+        Check if a publication associated with the pre_migrated distributor needs to be recreated.
+
+        No publications in the Container plugin.
+
+        Args:
+            pulp2distributor(Pulp2Distributor): Pre-migrated pulp2 distributor to check
+
+        Return:
+            bool: True, if a publication needs to be recreated; False if no changes are needed
+        """
+        return False
+
+    @classmethod
+    def needs_new_distribution(cls, pulp2distributor):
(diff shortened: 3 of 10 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -232,7 +232,8 @@ def get_remote_by_importer_id(importer_id):
             lazy_wo_catalog = is_lazy_type and not pulp2lazycatalog
             if pulp2content.pulp3_content is not None and (not is_lazy_type or lazy_wo_catalog):
                 if pb:
-                    pb.increment()
+                    pb.total -= 1
+                    pb.save()
```
goosemania, line 236: This is for the sake of not showing counters for the skipped content, otherwise, when you remigrate with no changes, it's the only counter for migration which is not 0.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -557,27 +579,55 @@ def mark_removed_resources(plan, type_to_repo_ids):
...
+            old_dist_query).only('pulp3_distribution')
+
+        for pulp2distributor in pulp2distributors_with_old_distributions_qs:
+            if pulp2distributor.is_migrated:
+                pulp2distributor.is_migrated = False
+                pulp2distributor.save()
+            if pulp2distributor.pulp3_publication:
+                pulp2distributor.pulp3_publication.delete()
+            if pulp2distributor.pulp3_distribution:
+                pulp2distributor.pulp3_distribution.delete()
```
goosemania, line 633: It's a pain that they are removed one by one.
I'm open for suggestion what to do here.

pulp_2to3_migration/app/migration.py
```diff
@@ -52,6 +52,8 @@ def migrate_repositories(plan):
     )
     with ProgressReport(**progress_data) as pb:
         for plugin in plan.get_plugin_plans():
+
+            # TODO: Should we filter out repos which have is_migrated=True?
```
dralley, line 56: Do these TODO's need to be addressed before merging?

goosemania: I'd prefer to but it's not necessary. They don't affect current implementation, it will work without them.
I just noticed few places where I was not sure why we are not checking the is_migrated and I wanted to run it by you and @ipanova

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -3,7 +3,10 @@
 from collections import namedtuple
 
 from django.db import transaction
-from django.db.models import Max
+from django.db.models import (
```
dralley, line 6: Nitpick: why split across 2 lines?

pulp_2to3_migration/app/plugin/iso/repository.py
```diff
@@ -72,3 +72,41 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
...
+        Check if a distribution associated with the pre_migrated distributor needs to be recreated.
+
+        Args:
+            pulp2distributor(Pulp2Distributor): Pre-migrated pulp2 distributor to check
+
+        Return:
+            bool: True, if a distribution needs to be recreated; False if no changes are needed
+
+        """
+        if not pulp2distributor.pulp3_distribution:
```
dralley, line 103: Right now it looks like this code is duplicated in 3 places. Can we make it the default implementation?

pulp_2to3_migration/app/pre_migration.py
dralley: ```select_related``` might work.

pulp_2to3_migration/app/plugin/content.py
dralley: +1

pulp_2to3_migration/app/plugin/iso/repository.py
goosemania: It's different in for the docker, because it has `repo-registry-id` instead of `relative_url`.

I tried to avoid default implementation on purpose, so there are no implications when new plugin is added, e.g. pulp_deb. The problem is that if the default logic is wrong for the plugin and a plugin writer is unaware of these methods and everything just works, distribution or publication might not be recreated when they potentially should be. So throwing the ImplementationError seemed to me like a good way to ensure plugin writers are conscious of what logic is applied here.

What do you think? I can be convinced that it's not the best way to handle that.

pulp_2to3_migration/app/pre_migration.py
goosemania: Just for the sake of keeping the style (if we have any). If there is more than one import, we split it across multiple lines. I can put it in one line, not a problem.

pulp_2to3_migration/app/migration.py
```diff
@@ -281,6 +285,7 @@ def create_repo_version(migrator, pulp3_repo_name, pulp2_repo, pulp3_remote=None
         pulp3_remote(remote): a pulp3 remote
     """
 
+    # TODO: adjust either the comment or the code due to tracking of remotes
```
ipanova, line 288: +1

pulp_2to3_migration/app/plugin/content.py
ipanova: +1, i meant to -1 instead of increment, that was my oversight.

pulp_2to3_migration/app/migration.py
ipanova: i think it should be the opposite, we should filter for is_migrated=True

ipanova: yes, let's set it to True, this way we will avoid additional step for repo_version_creation

pulp_2to3_migration/app/migration.py
```diff
@@ -191,6 +193,7 @@ def complex_repo_migration(plugin, pulp3_repo_setup, repo_name):
             pass
     for pulp2_repo_info in repo_versions_setup:
         try:
+            # TODO: should we check the is_migrated=True?
             pulp2_repo = Pulp2Repository.objects.get(
```
ipanova, line 197: same here, is_migrated=False

ipanova: and one line 216 is_migrated=True
`migrated_repo = Pulp2Repository.objects.get(pulp2_repo_id=repo_id, is_migrated=True, not_in_plan=False)`

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -457,16 +461,34 @@ def pre_migrate_distributor(repo_id, distributors, distributor_types, repo=None)
...
+
+        if not created:
+            # if it was marked as such because it was not present in the migration plan
+            distributor.not_in_plan = False
+
+            if last_updated != distributor.pulp2_last_updated:
+                distributor.pulp2_config = dist_data.config
+                dist_migrator = distributor_migrators.get(distributor.pulp2_type_id)
+                needs_new_publication = dist_migrator.needs_new_publication(distributor)
+                if needs_new_publication and distributor.pulp3_publication:
```
ipanova, line 482: for case when checksum type changes and relative_url changes as well, should not we remove both publication and distribution? meaning that` is else` will not be enough to cover this case

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -557,27 +579,55 @@ def mark_removed_resources(plan, type_to_repo_ids):
...
+
+        Pulp2Distributor.objects.bulk_update(objs=removed_dists,
+                                             fields=['not_in_plan'],
+                                             batch_size=1000)
+
 
 def delete_old_resources(plan):
     """
-    Delete pre-migrated and migrated data for resources which are migrated fully on every run.
+    Delete old Publications/Distributions which are no longer present in Pulp2.
```
ipanova, line 603: should we do same for importers?

pulp_2to3_migration/app/pre_migration.py
dralley: Ah, does black require that now?

pulp_2to3_migration/app/plugin/iso/repository.py
dralley: ```ImplementationError``` sounds fine and is probably the best thing to do if we can't have a default.  I'm on the fence whether we should just have duplicate code or try to make it into a utility function kind of thing -- but this is fine with me if you have a preference for this way.

goosemania: No preference, utility function sounds like a fine idea.

pulp_2to3_migration/app/pre_migration.py
goosemania: Not sure, have not checked that

pulp_2to3_migration/app/pre_migration.py
goosemania: Not essential and there will be other orphaned resources, like old repo versions or repos themselves.

pulp_2to3_migration/app/pre_migration.py
goosemania: yeah, we need to always remove distributors if a publication changes

pulp_2to3_migration/app/migration.py
goosemania: Discussed that we don't need that, the create_repo_version doesn't do any extra work if a repo is migrated but it ensures that all remotes are set properly.

pulp_2to3_migration/app/migration.py
goosemania: Agreed to remove the comment

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -457,16 +461,34 @@ def pre_migrate_distributor(repo_id, distributors, distributor_types, repo=None)
...
+                dist_migrator = distributor_migrators.get(distributor.pulp2_type_id)
+                needs_new_publication = dist_migrator.needs_new_publication(distributor)
+                if needs_new_publication and distributor.pulp3_publication:
+                    distributor.pulp3_publication.delete()
+                else:
+                    needs_new_distribution = dist_migrator.needs_new_distribution(distributor)
+                    if needs_new_distribution and distributor.pulp3_distribution:
+                        distributor.pulp3_distribution.delete()
+
+                distributor.pulp2_last_updated = last_updated
```
ipanova, line 489: we should update old config with the new one
distributor.pulp2_config = dist_data.config

ipanova: nevermind, i see it on line 479, maybe move all field updates in one place for better readability?

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -457,16 +461,34 @@ def pre_migrate_distributor(repo_id, distributors, distributor_types, repo=None)
...
+        if not created:
+            # if it was marked as such because it was not present in the migration plan
+            distributor.not_in_plan = False
+
+            if last_updated != distributor.pulp2_last_updated:
+                distributor.pulp2_config = dist_data.config
+                dist_migrator = distributor_migrators.get(distributor.pulp2_type_id)
+                needs_new_publication = dist_migrator.needs_new_publication(distributor)
+                if needs_new_publication and distributor.pulp3_publication:
+                    distributor.pulp3_publication.delete()
```
ipanova, line 483: i suggest adding
```
if needs_new_publication and distributor.pulp3_publication:
    distributor.pulp3_publication.delete()
    if distributor.pulp3_distribution:
        distributor.pulp3_distribution.delete()
```

pulp_2to3_migration/app/plugin/docker/repository.py
```diff
@@ -57,3 +57,41 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
         base_config['repository_version'] = repo_version
         distribution, created = ContainerDistribution.objects.update_or_create(**base_config)
```
ipanova, line 58: this needs to be changed to:
`update_or_create(name=name, base_path=base_path, defaults=base_config)`

pulp_2to3_migration/app/pre_migration.py
dralley: I think I misunderstood what ```select_related``` does.  Anyway, something like this might work:

```
pulp2distributors_with_old_distributions_qs.bulk_update(is_migrated=True)
pulp2distributors_with_old_distributions_qs.pulp3_publication_set.delete()
pulp2distributors_with_old_distributions_qs.pulp3_distribution_set.delete()
```

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -557,27 +579,55 @@ def mark_removed_resources(plan, type_to_repo_ids):
                                           fields=['not_in_plan'],
                                           batch_size=1000)
 
+        # Mark distributors
+        mongo_dist_object_ids = set(str(i.id) for i in Distributor.objects.only('id'))
```
dralley, line 583: Adding ```.as_pymongo()``` here would give you something equivalent to ```.values_list()```

pulp_2to3_migration/app/pre_migration.py
goosemania: Yeah, I moved the config up because it is used in the calls right after it.
Thanks, I moved everything up.

PR #154, opened 2020-05-28
Consider all repos present in the plan, not only the ones to migrate.

goosemania:
We need to go through all repos mentioned in the plan
in case importers' or distributors' repos are not being migrated.
Otherwise, such importers and distributors are ignored and not migrated.

closes #6853
https://pulp.plan.io/issues/6853


dralley:
Needs changelog entry

PR #170, opened 2020-06-23
Make size fields of BigIntegerField type.

goosemania:
To be inline with pulpocre `Artifact.size`.

closes #6939
https://pulp.plan.io/issues/6939


ipanova:
@goosemania will you add a note to the plugin writers guide?

PR #173, opened 2020-07-02
Fix RPM migration when its remote is not migrated.

goosemania:
closes #7078
https://pulp.plan.io/issues/7078


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -372,7 +373,8 @@ def get_remote_by_importer_id(importer_id):
                 pb.increment()
 
             if has_future:
```
ipanova, line 375: if has_future and dc:

goosemania: that was my first thought but then if has_future=True but there is no dc, a batch of futures might not be resolved.

PR #176, opened 2020-07-08
Fix yum metadata file not being migrated

dralley:
closes: #7093
https://pulp.plan.io/issues/7093

### DIFF
--- a/pulp_2to3_migration/app/plugin/rpm/migrator.py
+++ b/pulp_2to3_migration/app/plugin/rpm/migrator.py
@@ -69,16 +69,17 @@ class RpmMigrator(Pulp2to3PluginMigrator):
     """
     pulp2_plugin = 'rpm'
     pulp2_content_models = {
+        'rpm': pulp2_models.RPM,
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/plugin/rpm/migrator.py
```diff
@@ -69,16 +69,17 @@ class RpmMigrator(Pulp2to3PluginMigrator):
     """
     pulp2_plugin = 'rpm'
     pulp2_content_models = {
+        'rpm': pulp2_models.RPM,
+        'srpm': pulp2_models.SRPM,
         'distribution': pulp2_models.Distribution,
         'erratum': pulp2_models.Errata,
         'modulemd': pulp2_models.Modulemd,
         'modulemd_defaults': pulp2_models.ModulemdDefaults,
+        'yum_repo_metadata_file': pulp2_models.YumMetadataFile,
```
dralley, line 78: All I added was this line, the rest is just re-ordering it to match up w/ the lines below, because the fact that they're in different orders makes it easier for bugs to be hidden.

PR #177, opened 2020-07-17
Fix a bug with multiple copies of an errata end up in a repo

dralley:
The packagelist of an errata depends on content within the repo. When an
errata is copied from one Pulp 2 repository to another, it is the same
content unit but with a different package list, but in Pulp 3 these
would be considered separate content units.

Because of this, when content is migrated, two Pulp 3 UpdateRecords are
created from the same Pulp2Content. Later when content units are
associated with the repositories, no accounting was being taken for the
fact that this could be the case, and that we needed to pick the
particular content unit of the form that existed in the Pulp 2
repository rather than all of them.ul

re: #7165
https://pulp.plan.io/issues/7165

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -350,8 +350,11 @@ def create_repo_version(migrator, progress_rv, pulp2_repo, pulp3_remote=None):
     pulp3_repo = pulp2_repo.pulp3_repository
     unit_ids = Pulp2RepoContent.objects.filter(pulp2_repository=pulp2_repo).values_list(
         'pulp2_unit_id', flat=True)
-    incoming_content = set(Pulp2Content.objects.filter(pulp2_id__in=unit_ids).only(
-        'pulp3_content').values_list('pulp3_content__pk', flat=True))
+    incoming_content = set(
+        Pulp2Content.objects.filter(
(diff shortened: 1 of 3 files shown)


dralley:
@goosemania Take a look at the second commit, I'll rebase it before merging.

pulp_2to3_migration/app/migration.py
```diff
@@ -350,8 +350,11 @@ def create_repo_version(migrator, progress_rv, pulp2_repo, pulp3_remote=None):
     pulp3_repo = pulp2_repo.pulp3_repository
     unit_ids = Pulp2RepoContent.objects.filter(pulp2_repository=pulp2_repo).values_list(
         'pulp2_unit_id', flat=True)
-    incoming_content = set(Pulp2Content.objects.filter(pulp2_id__in=unit_ids).only(
-        'pulp3_content').values_list('pulp3_content__pk', flat=True))
+    incoming_content = set(
+        Pulp2Content.objects.filter(
+            Q(pulp2_id__in=unit_ids) & (Q(pulp2_repo=None) | Q(pulp2_repo=pulp2_repo)),
```
goosemania, line 355: Nice catch :+1:   It's only advisories for which pulp2repo is set.

PR #178, opened 2020-07-20
Fix a bug where errata were not always migrated for new repositories

dralley:
If an errata has already been migrated and is then added to a new
repository, which is then migrated, the new repository does not contain
the affected errata.

Ensure that new Pulp2Content are created when existing content are
migrated into new repos for the errata corner case (needs one per each repo).

closes: #7092
https://pulp.plan.io/issues/7092

### DIFF
--- a/pulp_2to3_migration/app/tasks/migrate.py
+++ b/pulp_2to3_migration/app/tasks/migrate.py
@@ -132,9 +132,9 @@ def migrate_from_pulp2(migration_plan_pk, validate=False, dry_run=False):
     # TODO: if plan is empty for a plugin, only migrate downloaded content
 
     pre_migrate_all_without_content(plan, type_to_repo_ids, repo_id_to_type)
+    pre_migrate_all_content(plan)
     mark_removed_resources(plan, type_to_repo_ids)
     delete_old_resources(plan)
-    pre_migrate_all_content(plan)
     migrate_repositories(plan)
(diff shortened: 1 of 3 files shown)


dralley:
Will rebase commits after review

pulp_2to3_migration/app/tasks/migrate.py
```diff
@@ -132,9 +132,9 @@ def get_repo_types(plan):
     # TODO: if plan is empty for a plugin, only migrate downloaded content
 
     pre_migrate_all_without_content(plan, type_to_repo_ids, repo_id_to_type)
+    pre_migrate_all_content(plan)
```
goosemania, line 135: @dralley , why did this move?

dralley: I noticed that on line 205 in the other file (expand the code a couple of times) it's marking repositories as not-migrated and I figured that should happen prior to these other steps.  I think I was having issues with repository versions being kept around that shouldn't have been or something along those lines and it stopped after flipping this ordering (although I can't rule out that it was a different change I made).

I can try to flip it back and see if anything goes wrong again

goosemania: I think we fixed pulp2, so the latest pulp2 version should not require this code block on line 205, but it needs to be tested...

PR #179, opened 2020-07-21
Fix high memory consumption issues

dralley:
closes: #7152
https://pulp.plan.io/issues/7152

### DIFF
--- a/CHANGES/7152.bugfix
+++ b/CHANGES/7152.bugfix
@@ -1,0 +1,1 @@
+Fix an issue causing extremely high memory usage as # of content scale up.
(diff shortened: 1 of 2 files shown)


dralley:
Before and after with a single repository: https://imgur.com/a/mrqAJ2k

With 8 repositories and ~230k units, it maxes out around 600mb of RAM.  That definitely would have run out of memory before in my experience, even with ~12gb allocated to the box.

I'm going to keep looking into this tomorrow because I want to make sure the tool I'm using isn't lying to me, since a lot of what we're doing is using C-based libraries that Python profiling tools can't always necessarily track.  Also, it's a lot more difficult to track the memory usage of many different processes than one process, so I made a few hacks to execute the task without forking and disable parallel repo migration, so I need some proper measurements from the system level with more of a real-world setup.

I'll also probably look into https://pulp.plan.io/issues/7183

goosemania:
:tada: :cat2: :1st_place_medal:

PR #194, opened 2020-07-27
Fix exceptions not bubbling up through the task

dralley:
closes: #6469
https://pulp.plan.io/issues/6469

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -181,8 +181,8 @@ class ContentMigrationFirstStage(Stage):
                 # We are waiting on the coroutine to finish, because the order of the processed
                 # content for plugins like Container and RPM is important because of the relations
                 # between the content types.
-                await asyncio.wait(
-                    [self.migrate_to_pulp3(pulp_2to3_detail_qs, self.migrator, ctype, pb=pb)]
+                await asyncio.gather(
+                    self.migrate_to_pulp3(pulp_2to3_detail_qs, self.migrator, ctype, pb=pb)
                 )
 
     async def migrate_to_pulp3(self, batch, migrator, content_type, pb=None):
--- a/CHANGES/6469.bugfix
+++ b/CHANGES/6469.bugfix
@@ -1,0 +1,1 @@
+Fix exceptions thrown by content migration not being bubbled up through the task.


dralley:
Apparently ```asyncio.wait()``` swallows exceptions (they become part of the returned tasks or something that you need to manually check) instead of raising.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -181,8 +181,8 @@ async def run(self):
                 # We are waiting on the coroutine to finish, because the order of the processed
                 # content for plugins like Container and RPM is important because of the relations
                 # between the content types.
-                await asyncio.wait(
-                    [self.migrate_to_pulp3(pulp_2to3_detail_qs, self.migrator, ctype, pb=pb)]
+                await asyncio.gather(
+                    self.migrate_to_pulp3(pulp_2to3_detail_qs, self.migrator, ctype, pb=pb)
```
dralley, line 185: And actually to be honest I'm not sure the ```gather()``` is necessary either.  This is a non-async for-loop so  ```await```ing here should intrinsically mean that it doesn't move on until the coroutine returns.

goosemania: I agree.
Does it make sense to make the async loop? Thoug the majority of the operations we have here are the db ones, so they are blocking anyway :/

dralley: I don't think it makes sense to use an async for loop

goosemania:
@dralley, I suggest to test on a small docker repo (since we no longer have dependent relations in RPM) before merging it, this will show if it's an ok adjustment because many things have changed since this  part was implemented. It will ensure that we haven't miscalculated/misinterpret anything. I believe @ipanova  spent a lot of time debugging stages because it was a hard issue to spot. Here is the corresponding commit  https://github.com/pulp/pulp-2to3-migration/commit/2875e39929983f74efbb81fd5d3fce1c6211ee15 which did exactly the inverse of what you are doing.

I think a small docker repo is a key because all content is in one batch and it can mess up things not deterministically.

dralley:
Good catch, in that case, maybe I should just leave it as was with ```gather()```.

@ipanova Do you happen to know if any other changes would have addressed this problem since February?

goosemania:
@dralley , `gather` doesn't seem to add a noticeable overhead for performance, so maybe it's safer to have it, in case we don't want to spend more time on this issue. Just a thought.

ipanova:
@dralley i don't think there was any change since then, relations are still a thing in docker and the order also matters.

I have ran a quick test without acync loop and i indeed have a failure
```
    "error": {
        "description": "A file failed validation due to checksum.",
        "traceback": "  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/worker.py\", line 886, in perform_job\n    rv = job.perform()\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py\", line 664, in perform\n    self._result = self._execute()\n  File \"/usr/local/lib/pulp/lib64/python3.6/site-packages/rq/job.py\", line 670, in _execute\n    return self.func(*self.args, **self.kwargs)\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/tasks/migrate.py\", line 140, in migrate_from_pulp2\n    migrate_content(plan)\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/migration.py\", line 36, in migrate_content\n    plugin.migrator.migrate_content_to_pulp3()\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/docker/migrator.py\", line 102, in migrate_content_to_pulp3\n    loop.run_until_complete(dm.create())\n  File \"/usr/lib64/python3.6/asyncio/base_events.py\", line 484, in run_until_complete\n    return future.result()\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/content.py\", line 84, in create\n    await pipeline\n  File \"/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py\", line 225, in create_pipeline\n    await asyncio.gather(*futures)\n  File \"/home/vagrant/devel/pulpcore/pulpcore/plugin/stages/api.py\", line 43, in __call__\n    await self.run()\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/content.py\", line 187, in run\n    await self.migrate_to_pulp3(pulp_2to3_detail_qs, self.migrator, ctype, pb=pb)\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/content.py\", line 335, in migrate_to_pulp3\n    downloaded=pulp2content.downloaded)\n  File \"/home/vagrant/devel/pulp-2to3-migration/pulp_2to3_migration/app/plugin/content.py\", line 121, in create_artifact\n    expected_size=expected_size)\n  File \"/home/vagrant/devel/pulpcore/pulpcore/app/models/content.py\", line 241, in init_and_validate\n    raise DigestValidationError()\n"

```

dralley:
@ipanova That is w/o either the "wait" or the "gather" correct?

ipanova:
> @ipanova That is w/o either the "wait" or the "gather" correct?



> @ipanova That is w/o either the "wait" or the "gather" correct?

correct. just `await self.migrate_to_pulp3(pulp2content_qs, pb=pb)`

PR #195, opened 2020-07-28
Add a teardown util which removes all data from pulp db.

goosemania:
Run it after each existing test.

closes #6963
https://pulp.plan.io/issues/6963

### DIFF
--- a/pulp_2to3_migration/tests/functional/util.py
+++ b/pulp_2to3_migration/tests/functional/util.py
@@ -1,5 +1,8 @@
 from time import sleep
 
+from django.core.management.color import no_style
+from django.db import DEFAULT_DB_ALIAS, connections
+
 
 def monitor_task(tasks_api, task_href):
     """Polls the Task API until the task is in a completed state.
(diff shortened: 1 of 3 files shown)


pulp_2to3_migration/tests/functional/util.py
```diff
@@ -24,3 +27,18 @@ def monitor_task(tasks_api, task_href):
...
+
+    Cleans all the Pulp 3 tables without shutting down the services.
+    """
+    database = DEFAULT_DB_ALIAS
+    connection = connections[database]
+    style = no_style()
+    tables = connection.introspection.django_table_names(only_existing=True, include_views=False)
+    tables.remove('auth_user')
+    sql_list = connection.ops.sql_flush(style, tables, sequences=())
+    connection.ops.execute_sql_flush(database, sql_list)
```
goosemania, line 44: I t was painful but it seems to work, however it requires pulp for introspection, so it will be a new dependency for functional tests which kinda devalues the approach of running tests on the machine outside of pulp.

I wonder if it's better to just perform SQL manually just by looking at the db schema and cleaning all the table there (except the auth group).

Thoughts? @dralley @dkliban

dralley: Does that mean it work work on Travis?  If that is not the case then, while not ideal, I'd consider it acceptable.  If it is a blocker then I agree we should find some other way of implementing it, such as checking the DB schema directly.

goosemania: @dralley , I think it means that we can't run it on ubuntu and would need to run functional tests in a container with fedora.

goosemania:
Closed in favour of https://github.com/pulp/pulp-2to3-migration/pull/208

PR #197, opened 2020-07-30
Assorted performance improvements

dralley:
[noissue]


dralley:
Pagination improves performance by about 30% vs. what it was after the iterator() change.  I haven't measured the memory usage again so I'm going to do that, and also look for other places of improvement.

dralley:
I haven't formally memory usage again yet (I will eventually) but I took glances at htop and it doesn't look like it becomes too significant

PR #198, opened 2020-07-31
Don't iterate all lazy content every time we re-migrate

dralley:
Push some of our Python-level checks into database query logic to
significantly reduce the # of individual queries we make and the amount
of content we need to iterate over / prefetch for.

[noissue]

### DIFF
--- a/CHANGES/6111.bugfix
+++ b/CHANGES/6111.bugfix
@@ -1,0 +1,1 @@
+Significantly improved performance of partial migrations (when some content / repos has been migrated already).
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -7,11 +7,11 @@ from urllib.parse import urljoin
 
 from gettext import gettext as _
 
+from cursor_pagination import CursorPaginator
 from django.conf import settings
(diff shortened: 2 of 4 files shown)


dralley:
Initial migration: 232 seconds  (no change)

Re-migration: 13 seconds (reduced from 33 seconds)

dralley:
Including the tweaks and the RPM publishing improvements from https://github.com/pulp/pulp_rpm/pull/1795, re-migration is down to 10 seconds, initial migration is down to 211 seconds.

### Before (re-migration)

original_migration_performance_remigration

### After (re-migration)

new_migration_performance_remigration

Most of the time is being spent in inserts now so I'm not sure how to further improve things.  It looks like Django ORM actually adds quite a bit of overhead here so there's still possibly room for improvement if we were to drop down a level and use the database driver directly or something like SQLAlchemy's query builder.

I'm not going to spend any time on that though. It's probably getting close to "good enough".   I'll work on initial migration performance a little bit (would also help re-migration when things are actually changing, which they are not in my tests) and then check our memory consumption again and call it done.

dralley:
My rewritten stages knock about 13% off of the initial migration, 211 seconds -> 170 seconds

PR #207, opened 2020-08-07
Update License Text to Match GPL-2.0

amayer5125:
Copy exact license text from [choosealicense.com](https://choosealicense.com/).
This will help GitHub recognize the license. Instead of displaying "View license" it will say "GPLv2". The license name will also show up in different places on GitHub such as in search results.
image


goosemania (approved):
Thanks for the patch!

PR #208, opened 2020-08-11
[Raw SQL] Add a teardown util which removes all data from pulp db.

goosemania:
Run it after each existing test.

closes #6963
https://pulp.plan.io/issues/6963

### DIFF
--- a/pulp_2to3_migration/tests/functional/dynaconf_config.py
+++ b/pulp_2to3_migration/tests/functional/dynaconf_config.py
@@ -1,0 +1,3 @@
+from dynaconf import Dynaconf
+
+settings = Dynaconf(settings_files=['func_test_settings.py', '/etc/pulp/settings.py'])
--- a/pulp_2to3_migration/tests/functional/util.py
+++ b/pulp_2to3_migration/tests/functional/util.py
@@ -1,28 +1,6 @@
-from contextlib import closing
 from time import sleep
 
-import psycopg2
-
 from .dynaconf_config import settings
-
-QUERY_ALL_TABLES = """
-    SELECT table_name FROM information_schema.tables
-        WHERE table_schema = 'public'
-"""
-TABLES_TO_KEEP = [
-    # django's sqlclear or sqlflush excludes this table when cleaning up the db
-    'django_migrations',
-
-    # not to create an admin user every time
-    'auth_user',
-
-    # not to be doomed by the lack of permissions
-    'auth_permission',
-    'core_accesspolicy',
-
-    # 'auth_permission' references it, so it should not be truncated
-    'django_content_type',
-]
 
 
 def monitor_task(tasks_api, task_href):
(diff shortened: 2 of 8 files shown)


goosemania:
An alternative to  https://github.com/pulp/pulp-2to3-migration/pull/195, not to require pulp being installed on the system for functional tests.

pulp_2to3_migration/tests/functional/dynaconf_config.py
```diff
@@ -0,0 +1,3 @@
+from dynaconf import Dynaconf
+
+settings = Dynaconf(settings_files=['/etc/pulp/settings.py'])
```
goosemania, line 3: I just realized that this requires pulp :)
Any suggestion where to have credentials for the DB (for functional tests)?
hardcode in the code? set in travis as env vars? anything else?
The downside of setting env vars in travis is that it makes it inconvenient to run tests locally.

dkliban: You don't actually have to have Pulp. You can just use Dynaconf and a settings file.

dkliban: Maybe have a separate settings file tests?

goosemania: That's what I did here, but Travis doesn't have /etc/pulp/settings.py, because pulp is not there.
Are you suggesting to have a settings file for the func tests? (which is a form of hardcoding, and it needs to be changed if we changed defaults in pulp)

goosemania: disregard my comment, I didn't see your question when replied

dralley: @goosemania If we don't want to put it in the same location, I suggest ```.config/pulp/settings.py```

goosemania: https://docs.pulpproject.org/installation/configuration.html#by-configuration-file

goosemania: Maybe this will work?

```diff
diff --git a/pulp_2to3_migration/tests/functional/dynaconf_config.py b/pulp_2to3_migration/tests/functional/dynaconf_config.py
index 8640a75..7653b14 100644
--- a/pulp_2to3_migration/tests/functional/dynaconf_config.py
+++ b/pulp_2to3_migration/tests/functional/dynaconf_config.py
@@ -1,3 +1,7 @@
 from dynaconf import Dynaconf
 
-settings = Dynaconf(settings_files=['/etc/pulp/settings.py'])
+try:
+    settings = Dynaconf(settings_files=['/etc/pulp/settings.py'])
+    settings.DATABASES
+except AttributeError:
+    settings = Dynaconf(settings_files=['~/.config/pulp/settings.py'])
```

dralley: I think that would work.  I wonder if there's a more elegant way to do fallback from one possible location to another, maybe by listing both files in the proper order in the original invocation?

But I would be fine with merging this as-is.

goosemania: My understanding is that 2 files specified there are getting merged if both are present which might not be desirable.

pulp_2to3_migration/tests/functional/util.py
```diff
@@ -24,3 +26,25 @@ def monitor_task(tasks_api, task_href):
     else:
         print("The task did not finish successfully.")
     return task
+
+
+def get_psql_smash_cmd(sql_statement):
```
dralley, line 31: Why return the command instead of executing immediately?

dralley: Semi-related, why execute from the command line vs. something like ```psycopg2```?

https://www.psycopg.org/docs/usage.html

goosemania: > Why return the command instead of executing immediately?

I tried to avoid instantiating smash cli client every time, I guess it can be passed here as an argument to execute right away.
To be clear what smash cli does - it literally runs the command inside the container `docker exec -it pulp psql ...`


> Semi-related, why execute from the command line vs. something like `psycopg2`?
> 
> https://www.psycopg.org/docs/usage.html

The problem with programmatic approach is the following:
 - func tests are running on the travis machine
 - the db is local in our dev environment and it's in the container in our CI
 - in order to connect from the host to the db in a container, the db should listen not only on localhost (which is doable but a bit of a pain if it's for the tests sake only) + the db port should be exposed (easily doable but yet another customisation)
 - x9c4 convinced me that smash cli (even though I really dislike it :)) is a better approach because you can configure different transport and have pulp in a container or on a different machine and smash cli will either run `docker exec` or ssh into a machine.

Suggestions are welcome.

PR #209, opened 2020-08-17
Update travis

goosemania:
[noissue]

### DIFF
--- a/.travis/publish_plugin_pypi.sh
+++ b/.travis/publish_plugin_pypi.sh
@@ -9,25 +9,7 @@
 
 set -euv
 
-export COMMIT_MSG=$(git log --format=%B --no-merges -1)
-export RELEASE=$(echo $COMMIT_MSG | awk '{print $2}')
-export MILESTONE_URL=$(echo $COMMIT_MSG | awk '{print $6}')
-export REDMINE_QUERY_URL=$(echo $COMMIT_MSG | awk '{print $4}')
-
-echo "Releasing $RELEASE"
-echo "Milestone URL: $MILESTONE_URL"
-echo "Query: $REDMINE_QUERY_URL"
-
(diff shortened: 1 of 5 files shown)


.travis/post_before_script.sh
```diff
@@ -16,7 +16,7 @@ mongo pulp_database --eval 'db.createUser({user:"travis",pwd:"travis",roles:["re
 
 # Otherwise, pulpcore will get installed as the stable release from PyPI
 # as a dep of pulp-2to3-migration.
-pip install $TRAVIS_BUILD_DIR/../pulpcore
+#pip install $TRAVIS_BUILD_DIR/../pulpcore
 
-pip install .
+#pip install .
```
goosemania, line 21: @dkliban , I'm trying to figure out why it needs to be installed outside of  the container, any hints?

pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -29,7 +29,7 @@
 
 
 # Initialize MongoDB connection
-initialize()
+# initialize()
```
goosemania, line 32: @dkliban @dralley , why do we need to initialize a mongodb connection from the functional tests code? How do we use it? I thought that Pulp which runs in a container does it when we trigger a task to migrate from pulp 2 to pulp 3.

goosemania: Found it. We use mongo to find the list of repositories in pulp2.

.travis/post_before_script.sh
goosemania: It seems that we need configs for functional tests  from both pulpcore settings and pulp-2to3-migration settings

ipanova (approved):

PR #211, opened 2020-08-17
Don't entirely delete mutable content and re-pre-migrate them every time

dralley:
closes: #7280
https://pulp.plan.io/issues/7280


dralley:
Given how many times I had to quit and start over in frustration, I kind of expect this is still buggy, so I'm going to keep working on top of it for a couple days.  I'm also not entirely convinced this is the correct place for this code to be located.

dralley:
@goosemania Let's get https://github.com/pulp/pulp-2to3-migration/pull/240 merged first since that issue was making it difficult to test this PR properly.

PR #214, opened 2020-08-20
Set pulpocre version to be used at release time.

goosemania:
[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -30,7 +30,7 @@ plugin_name: pulp-2to3-migration
 plugin_snake: pulp_2to3_migration
 pulp_settings: null
 pulpcore_branch: 3.4
-pulpcore_pip_version_specifier: null
+pulpcore_pip_version_specifier: 3.4.1
 pydocstyle: false
 pypi_username: pulp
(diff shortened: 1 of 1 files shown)


ipanova (approved):

template_config.yml
```diff
@@ -30,7 +30,7 @@ plugin_name: pulp-2to3-migration
 plugin_snake: pulp_2to3_migration
 pulp_settings: null
 pulpcore_branch: 3.4
-pulpcore_pip_version_specifier: null
+pulpcore_pip_version_specifier: 3.4.1
```
dralley, line 33: Unfamiliar with this template option, does this basically lock the plugin to this version of pulpcore?

goosemania: @dralley , I believe (at least it's my intention here) that when travis runs release jobs (`is tag present` condition is satisfied) the installer will use this specific pulpcore version, otherwise it will use the latest.
Since 0.2 is compatible with 3.4, we need to set it. (I did not and my tag job failed on compatibility error).

The only thing that locks the plugin on the specific version is the requirements.txt I think.

PR #215, opened 2020-08-20
Update travis to use pulpcore 3.4.1

goosemania:
[noissue]


.travis/install.sh
```diff
@@ -71,7 +71,7 @@ image:
   tag: "${TAG}"
 plugins:
   - name: pulpcore
-    source: pulpcore
+    source: pulpcore3.4.1
```
goosemania, line 74: Is it correct that there is no space between the version number and `pulpcore`?

PR #233, opened 2020-08-28
Do not set is_migrated=False for LCE when importer is being re-migrated.

goosemania:
closes #7418
https://pulp.plan.io/issues/7418

Some already migrated LCE might not exist in Pulp 2 anymore, trying to migrate those causes issues.
Identifying old ones is not easy, because htere are no timestamps in pulp 2 for lazy catalog.

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -425,13 +425,8 @@ def pre_migrate_importer(repo_id, importers, importer_types, repo=None):
             if importer.pulp2_config.get('feed') != importer_data.config.get('feed'):
                 importer.pulp3_remote.delete()
                 importer.pulp3_remote = None
-                # find LCEs
-                pulp2lazycatalog = Pulp2LazyCatalog.objects.filter(
-                    pulp2_importer_id=importer.pulp2_object_id)
-                for lce in pulp2lazycatalog:
-                    lce.is_migrated = False
-                Pulp2LazyCatalog.objects.bulk_update(objs=pulp2lazycatalog,
-                                                     fields=['is_migrated'])
+                # do not flip is_migrated to False for LCE for at least once migrated importer
+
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -425,13 +425,8 @@ def pre_migrate_importer(repo_id, importers, importer_types, repo=None):
...
                 importer.pulp3_remote.delete()
                 importer.pulp3_remote = None
-                # find LCEs
-                pulp2lazycatalog = Pulp2LazyCatalog.objects.filter(
-                    pulp2_importer_id=importer.pulp2_object_id)
-                for lce in pulp2lazycatalog:
-                    lce.is_migrated = False
-                Pulp2LazyCatalog.objects.bulk_update(objs=pulp2lazycatalog,
-                                                     fields=['is_migrated'])
+                # do not flip is_migrated to False for LCE for at least once migrated importer
```
dralley, line 428: This will effectively just leave a bunch of 'orphaned' LCEs correct?  We won't have a problem with that?

goosemania: @dralley , yeah, "orphaned" Pulp2LazyCatalog entries in a sense that they are no longer present in pulp2.
If they are marked as migrated, they won't give us any problem, except the increased size of the table.
There will be no orphaned RemoteArtifacts in Pulp 3, they are removed when a remote is removed.

We can't clean Pulp2LazyCatalog entries up easily, because we have no timestamp. To clean them, our options are:
 - make a diff between pulp 2 and premigrated Pulp2LazyCatalog
 - or remove all and re-migrate always all lazy catalog entries into RemoteArtifacts (a lot of work)
 - change a unique constraint so we always have only one LCE for each artifact (it might work for some case, it might be a problem for others, e.g. we picked too old LCE with outdated URL )
 - any other option I missed

ipanova (approved):

PR #234, opened 2020-09-01
Fix a distribution tree LCE migration from a recreated pulp2 repo.

goosemania:
closes #7080
https://pulp.plan.io/issues/7080

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -283,6 +283,8 @@ class ContentMigrationFirstStage(Stage):
                     # find already created pulp3 content
                     pulp3content = pulp2content.pulp3_content
                     extra_info = None
+                    if is_multi_artifact:
+                        extra_info = pulp_2to3_detail_content.get_treeinfo_serialized()
 
                 else:
                     # create pulp3 content and assign relations if present
--- a/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
+++ b/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
@@ -723,9 +723,13 @@ class Pulp2Distribution(Pulp2to3Content):
             for distribution in pulp2_distribution_content_batch]
         cls.objects.bulk_create(pulp2distribution_to_save, ignore_conflicts=True)
 
-    def create_pulp3_content(self):
+    def get_treeinfo_serialized(self):
         """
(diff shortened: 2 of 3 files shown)


goosemania:
Tested by Katello

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -283,6 +283,8 @@ def get_remote_by_importer_id(importer_id):
                     # find already created pulp3 content
                     pulp3content = pulp2content.pulp3_content
                     extra_info = None
+                    if is_multi_artifact:
+                        extra_info = pulp_2to3_detail_content.get_treeinfo_serialized()
```
ipanova, line 287: what about moving this line after https://github.com/pulp/pulp-2to3-migration/blob/5eaba43c5ffb414acf678c037a410aca0af0db9d/pulp_2to3_migration/app/plugin/content.py#L296

dralley: I think the problem is line 293. Might be some way to rearrange it though.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -723,9 +723,13 @@ def pre_migrate_content_detail(cls, content_batch):
             for distribution in pulp2_distribution_content_batch]
         cls.objects.bulk_create(pulp2distribution_to_save, ignore_conflicts=True)
 
-    def create_pulp3_content(self):
+    def get_treeinfo_serialized(self):
```
dralley, line 726: This is just a refactor right, I'm not sure I see any functional changes in this section?

goosemania: The goal was to avoid instantiating a DistributionTree object if I need just a serialized metadata. So it is a refactor without logic changes in this place.

pulp_2to3_migration/app/plugin/content.py
ipanova: right

ipanova (approved):

jlsherrill:
this resolved the issue for me!

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
dralley: +1

PR #240, opened 2020-09-23
Fix migration of lazy multi-artifact content not present in plan

dralley:
closes: #7562
https://pulp.plan.io/issues/7562

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -271,7 +271,9 @@ class ContentMigrationFirstStage(Stage):
                 if is_lazy_type:
                     # get all Lazy Catalog Entries (LCEs) for this content
                     pulp2lazycatalog = Pulp2LazyCatalog.objects.filter(
-                        pulp2_unit_id=pulp2content.pulp2_id, is_migrated=False)
+                        pulp2_unit_id=pulp2content.pulp2_id,
+                        is_migrated=False,
+                    )
 
                 if is_lazy_type and not pulp2content.downloaded and not pulp2lazycatalog:
                     _logger.warn(_('On_demand content cannot be migrated without an entry in the '
@@ -310,8 +312,14 @@ class ContentMigrationFirstStage(Stage):
 
                         lces = pulp2lazycatalog.filter(pulp2_storage_path=image_path)
                         if lces:
+                            remote_declarative_artifacts = []
+
                             for lce in lces:
                                 remote = get_remote_by_importer_id(lce.pulp2_importer_id)
+
+                                if not remote and not downloaded:
+                                    continue
+
                                 remotes.add(remote)
                                 da = DeclarativeArtifact(
                                     artifact=artifact,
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -310,8 +312,17 @@ def get_remote_by_importer_id(importer_id):
...
+                            # To create multiple remote artifacts, create multiple instances of
+                            # declarative content which will differ by url/remote in their
+                            # declarative artifacts
+                            at_least_one_lce_migrated = False
+
                             for lce in lces:
                                 remote = get_remote_by_importer_id(lce.pulp2_importer_id)
+
+                                if not remote and not downloaded:
+                                    continue
```
dralley, line 324: @goosemania I'm not 100% sure this is the correct fix?  I just copied the general strategy being used a couple lines down.

ipanova: it seems to be a reasonable change. however i cannot wrap my head around - can this logic lead to a half migrated unit because of multiple remotes and some of them can be missing?

ipanova:
@dralley here is a similar bug with the same error https://pulp.plan.io/issues/7405 although the root cause is different. should we try addressing it in this PR as well?

pulp_2to3_migration/app/plugin/content.py
dralley: I'm attempting to avoid that with line 330 and 339

dralley:
@ipanova Yes, I think so

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -310,16 +312,31 @@ def get_remote_by_importer_id(importer_id):
...
                         if lces:
+                            remote_declarative_artifacts = []
+
                             for lce in lces:
                                 remote = get_remote_by_importer_id(lce.pulp2_importer_id)
+
+                                if not remote and not downloaded:
+                                    continue
+
                                 remotes.add(remote)
```
goosemania, line 323: Maybe this should be under `if remote` condition, to cover the case when a remote is not present but the item is downloaded. What do you think? to avoid have a set with None value in it and create Declarative artifact with it later.

pulp_2to3_migration/app/plugin/content.py
goosemania: @dralley , +1 to approach, seems reasonable to me.
@ipanova , I think even if some remotes are missing, it's not a big issue if we have at least one successful lce migration.

goosemania (approved):

pulp_2to3_migration/app/plugin/content.py
ipanova: sounds good, thanks for confirmation

pulp_2to3_migration/app/plugin/content.py
dralley: Agree

dralley:
@ipanova I can't seem to reproduce https://pulp.plan.io/issues/7405 using many different repositories, even with that commit reverted and using other, older commits.  I'm going to close it and we can reopen it if we later run into it with a particular repo or situation.

PR #241, opened 2020-10-07
Release 0.4.0

goosemania:


requirements.txt
```diff
@@ -1,5 +1,5 @@
 django-cursor-pagination
-pulpcore>=3.6
+pulpcore>=3.6,<3.9
```
ipanova: should not it be >=3.7,<3.9

goosemania: The plugin is still compatible with 3.6, there were no incompatible changes for it in pulpcore 3.7 I believe.

ipanova (approved):

PR #244, opened 2020-10-07
Don't entirely delete mutable content and re-pre-migrate them every time

dralley:
closes: #7280
https://pulp.plan.io/issues/7280


dralley:
@goosemania @ipanova This is ready to be looked at again.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -76,29 +77,24 @@ def pre_migrate_content(content_model, mutable_type, lazy_type, premigrate_hook)
...
     _logger.debug('The latest migrated {type} content has {timestamp} timestamp.'.format(
         type=content_type,
         timestamp=last_updated))
-    set_pulp2_repo = content_model.pulp_2to3_detail.set_pulp2_repo
 
     query_args = {}
-
-    if not set_pulp2_repo:
-        # query only newly created/updated items
-        query_args["_last_updated__gte"] = last_updated
```
dralley, line 95: Problem A:  We only want to save ```Pulp2Content``` for new content (```_last_updated__gte=last_updated```)
Problem B:  But if  ```set_pulp2_repo``` is true for the content type, then we do want to save new ```Pulp2Content``` depending on whether the content was added to a new repo (referencing the RepoContent table)

Solution:  Make a completely separate loop to handle Problem B instead of trying to do both in the same piece of code.

goosemania:
It looks great!
I will test it tomorrow on a customer db.

One question, how does null values behave in comparisons like this `pulp2_created__gte=last_updated`, since `pulp2_created`can be null.

dralley:
@goosemania Null values are treated as "less than" by the filter.  So if I do a migration on master branch to populate the DB, then run django migration, all of the existing ```Pulp2RepoContent``` with empty values in those fields are not included by this query which I assume is the one you're referring to:

```
        content_relations = Pulp2RepoContent.objects.filter(
            pulp2_content_type_id=record._content_type_id,
            pulp2_repository__not_in_plan=False,
            pulp2_created__gte=last_updated
        ).select_related(
            'pulp2_repository'
        ).only(
            'pulp2_repository'
        )
```

I think that is the correct behavior, unless there's a problem or edge case that I'm missing.

goosemania:
@dralley , great! I was just not sure how `null` is compared, thanks.

goosemania:
Works like a charm!
I didn't manage to time it, but regardless it's a great improvement! Thank you so much!

goosemania (approved):

PR #248, opened 2020-10-12
 Use publish task from pulp_file to create FilePublications.

dkliban:
fixes: #7693
https://pulp.plan.io/issues/7693

### DIFF
--- a/pulp_2to3_migration/app/plugin/iso/repository.py
+++ b/pulp_2to3_migration/app/plugin/iso/repository.py
@@ -5,12 +5,7 @@ from pulp_2to3_migration.app.plugin.api import (
 )
 
 from pulp_file.app.models import FileRemote, FilePublication, FileDistribution
-from pulp_file.manifest import Manifest
-from pulp_file.app.tasks.publishing import populate
-
-from django.core.files import File
-
-from pulpcore.plugin.models import PublishedMetadata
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/plugin/iso/repository.py
```diff
@@ -69,12 +70,13 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
...
-                PublishedMetadata.create_from_file(
-                    file=File(open(manifest.relative_path, "rb")), publication=publication
-                )
+            with WorkingDirectory():
+                with FilePublication.create(repo_version, pass_through=True) as publication:
+                    manifest = Manifest('PULP_MANIFEST')
+                    manifest.write(populate(publication))
+                    PublishedMetadata.create_from_file(
+                        file=File(open(manifest.relative_path, "rb")), publication=publication
+                    )
```
goosemania, line 79: Does it make sense to just call `publish` https://github.com/pulp/pulp_file/blob/master/pulp_file/app/tasks/publishing.py#L17?

Great catch btw.

dkliban: It does make sense to do that. The only difference in the code is the logging.

goosemania (approved):

PR #249, opened 2020-10-12
Stop pre-migrating repository relations when a repo hasn't changed.

goosemania:
closes #7694
https://pulp.plan.io/issues/7694

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -382,8 +382,6 @@ def pre_migrate_all_without_content(plan, type_to_repo_ids, repo_id_to_type):
                     if not repos or repos and distributors_repos:
                         pre_migrate_distributor(
                             repo_id, distributors_repos, distributor_migrators, repo)
-                    if repo:
-                        pre_migrate_repocontent(repo)
                 pb.increment()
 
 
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -417,15 +417,17 @@ def pre_migrate_repo(record, repo_id_to_type):
...
+        if is_changed:
             repo.pulp2_last_unit_added = last_unit_added
             repo.last_unit_removed = last_unit_removed
             repo.pulp2_description = record.description
             repo.is_migrated = False
         repo.save()
 
-    return repo
+    if created or is_changed:
+        return repo
```
goosemania, line 430: It affects this https://github.com/pulp/pulp-2to3-migration/pull/249/files#diff-cb7ce2efbf29dbc7f500034642028cb9L385-L386.

I wonder if it makes sense to move the repocontent pre-migration here.

CHANGES/7694.feature
```diff
@@ -0,0 +1 @@
+Sped up repository pre-migration by skipping the relations pre-migration when nothing changed in a repository.
```
ipanova, line 1: s/ relations/ repository content relations

ipanova: i would probably use a .bugfix here or .misc, it does not bring new functionality

goosemania: Interesting :), I think about it from the other side. It's not a bug fix, it's a perf improvement :) and users should be aware of it, so I suggest not to use .misc.
I can convert it to a bug fix.

ipanova (approved):

CHANGES/7694.feature
ipanova: for the visibility reasons I prefer to convert it to the bugfix. I don't think any our previous perf improvements went as .feature :)

dralley: Actually, most of the previous performance improvements didn't come with changelog entries at all :)  Except when it was caused by an actual bug with user-facing impacts outside of performance.

I'm fine with all of the above including .misc.  Do you have a ballpark % of how much it improved things in your testing?

goosemania: @dralley , we should advertise perf improvements as much as we can. This is what I heard from @melcorr :) So let's try to make it visible in the changelog, wdyt?

Performance improvement here is for a specific case when a migration has been interrupted. Which we expect users to do if they migrate a large system. Those repositories which have been pre-migrated but not migrated yet, will no longer be pre-migrated one more time. In such cases pre-migration part went from 20-40 mins to 1-2 min, depending on the hardware.

If all repos has been migrated and nothing changed, this patch doesn't improve performance because we check if repo.is_migrated before pre-migrating it again.

dralley: > In such cases pre-migration part went from 20-40 mins to 1-2 min, depending on the hardware.

Cool, in that case it's definitely worth advertising.

PR #250, opened 2020-10-13
Optimize content migration queries to avoid sorts

dralley:
* Use select_related instead of prefetch_related to get everything in
   one query
* Use .iterator() instead of chunked_iterator() which required an
   additional sort operation
    
The result is not much faster on my dev machine, but much faster on the
Katello box w/ HDD.
    
closes: #7699
https://pulp.plan.io/issues/7699

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -7,11 +7,10 @@ from urllib.parse import urljoin
 
 from gettext import gettext as _
 
-from cursor_pagination import CursorPaginator
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
-from django.db.models import Prefetch, Q
+from django.db.models import Q
 
 from pulpcore.app.models import storage
 from pulpcore.plugin.models import (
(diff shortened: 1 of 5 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -89,30 +88,6 @@ async def create(self):
         await pipeline
 
 
-def chunked_queryset_iterator(queryset, size, *, ordering=('pk',)):
```
dralley, line 92: This is where the sort was introduced.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -251,22 +226,16 @@ def get_remote_by_importer_id(importer_id):
...
-            # We're using the PK which is not an incrementing integer so that doesn't hold
-            # true. However, at this point, all records have already been saved so we are safe.
-            chunked_iterator = chunked_queryset_iterator(
-                pulp_2to3_detail_qs.prefetch_related(*prefetch_args),
-                2000
-            )
+                select_extra.append('pulp2content__pulp2_repo')
 
-            for pulp_2to3_detail_content in chunked_iterator:
+            pulp_2to3_detail_qs = pulp_2to3_detail_qs.select_related(*select_extra)
```
dralley, line 237: I'm honestly not sure why I didn't write it like this originally.  Maybe I thought ```.select_related()``` also couldn't be used with ```.iterator()``` when it definitely can.

goosemania (approved):
Great improvement, thanks! :tada: :octocat:

PR #256, opened 2020-10-14
Small fixups

dralley:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -218,7 +218,7 @@ class ContentMigrationFirstStage(Stage):
 
         # order by pulp2_repo if it's set
         if content_model.set_pulp2_repo:
-            pulp_2to3_detail_qs = pulp_2to3_detail_qs.order_by('repo_id', 'pk')
+            pulp_2to3_detail_qs = pulp_2to3_detail_qs.order_by('repo_id')
 
         with ProgressReport(
             message='Migrating {} content to Pulp 3 {}'.format(self.migrator.pulp2_plugin,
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -235,7 +235,7 @@ def get_remote_by_importer_id(importer_id):
                 select_extra.append('pulp2content__pulp2_repo')
 
             pulp_2to3_detail_qs = pulp_2to3_detail_qs.select_related(*select_extra)
-            for pulp_2to3_detail_content in pulp_2to3_detail_qs.iterator():
+            for pulp_2to3_detail_content in pulp_2to3_detail_qs.iterator(chunk_size=800):
```
ipanova, line 238: out of curiosity, why 800?

ipanova (approved):

ipanova:
the build was failing for weird reasons so i restarted it.

pulp_2to3_migration/app/plugin/content.py
dralley: No particular reason, other than that it's smaller than 2000 but big enough to be worthwhile.  @goosemania was seeing some memory consumption issues and we were speculating about the causes - this might *not* be the problem, but it's worth trying since it's so easy.

ipanova: sounds good, I am probably biased by preferring having round number like 500,1000,2000,5000, etc <--- this is not a technically backed answer lol

dralley: It's as equally technically-backed as mine :)

goosemania:
Let's merge it, it works.

PR #262, opened 2020-10-29
Clear a dict which is not gc'ed and causes high memory usage.

goosemania:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
+++ b/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
@@ -150,7 +150,9 @@ class Pulp2RpmBase(Pulp2to3Content):
         cr_package = get_cr_obj(self)
         pkg_dict = Package.createrepo_to_dict(cr_package)
         pkg_dict['is_modular'] = self.is_modular
-        return (Package(**pkg_dict), None)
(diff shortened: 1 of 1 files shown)


pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -150,7 +150,9 @@ def create_pulp3_content(self):
         cr_package = get_cr_obj(self)
         pkg_dict = Package.createrepo_to_dict(cr_package)
         pkg_dict['is_modular'] = self.is_modular
-        return (Package(**pkg_dict), None)
+        package = Package(**pkg_dict)
+        pkg_dict.clear()
```
goosemania, line 154: So profiler showed that `Package.createrepo_to_dict` method in RPM plugin is using a lot of memory.
That method is just one big return of a dict. 

My guess is that pkg_dict is not getting garbage collected because it's an argument in return statement. Passing an argument is a reference count. Why it is not counted properly, I don't know.

I'm testing on a large system now to see that it indeed help. On a small system there was a noticeable improvement and profiler didn't show any accumulation of memory in one specific place.

Any more precise explanations are welcome.

goosemania:
Not the case.

PR #263, opened 2020-11-04
Add a reset/ endpoint to be able to run migration from scratch.

goosemania:
closes #7714
https://pulp.plan.io/issues/7714


pulp_2to3_migration/app/tasks/reset.py
```diff
@@ -0,0 +1,73 @@
...
+
+            repo_model = plugin.migrator.pulp3_repository
+            if hasattr(repo_model, 'sub_repo'):
+                # sub_repos can't be deleted until the content referring to them is not removed
+                repo_model.objects.filter(sub_repo=False).only('pk').delete()
+            else:
+                repo_model.objects.all().only('pk').delete()
+
+            for content_model in repo_model.CONTENT_TYPES:
+                content_model.objects.all().only('pk').delete()
```
goosemania, line 54: Any suggestions apart from the raw sql to do it differently?

goosemania: It would also be preferable to do it in batches because the memory consumption increases with the amount of content.
For 300 000 rpms, it was 2.6GB. Doesn't seem too bad but still...
To clean up (selectively) data for the large system (1K repos, 300K packages) takes a noticeable time, 30-40 mins.
I'm tempted to write sql, at least for the parts which do not have relations in the tables where data needs to be partially preserved.

dralley: Unless you explicitly singled out packages in your testing and if this is the same set of repos that contain 600k errata, I would guess that the errata are what takes so long and uses so much memory.  The reason being that Django has to handle the "cascade" deletion, and errata have a lot of object nesting with foreign keys to traverse.

goosemania: 600K are errata relations in pulp 2, we count it that way because we need to go through every single one relation and process pkglist for it.
The final number of advisories in pulp 3 is around 35K.
I believe packages are the ones to blame (for memory at least).

goosemania:
I heard from few folks, including katello, that it's acceptable to have this task running for 30-40 mins.

goosemania:
This implementation leaves Artifacts in place. Only Content and ContentArtifacts are getting removed.

pulp_2to3_migration/app/tasks/reset.py
```diff
@@ -0,0 +1,73 @@
...
+    with ProgressReport(**pb_data) as pb:
+        for plugin in plan.get_plugin_plans():
+            for pulp2to3_content_model in plugin.migrator.content_models.values():
+                pulp2to3_content_model.objects.all().only('pk').delete()
+        Pulp2Content.objects.all().only('pk').delete()
+        Pulp2Distributor.objects.all().only('pk').delete()
+        Pulp2Importer.objects.all().only('pk').delete()
+        Pulp2LazyCatalog.objects.all().only('pk').delete()
+        Pulp2RepoContent.objects.all().only('pk').delete()
+        Pulp2Repository.objects.all().only('pk').delete()
```
goosemania, line 72: My testing showed that I need to remove here by type and add indices.

docs/workflows.rst
```diff
@@ -64,3 +67,59 @@ All the commands should be run on Pulp 3 machine.
             }
         ]
     }
+
+
+Reset migrated Pulp 3 data
+--------------------------
+
+.. note::
+    Nothing is getting reset or modified in any way in Pulp 2.
```
dralley, line 76: s/is getting/will be/

docs/workflows.rst
```diff
@@ -64,3 +67,59 @@ All the commands should be run on Pulp 3 machine.
...
+
+.. note::
+    Migrated artifacts are untouched. If you are sure, you want to remove them, please run the
```
dralley, line 79: s/If you are sure, you want to remove them/If you are sure you want to remove them/

docs/workflows.rst
```diff
@@ -64,3 +67,59 @@ All the commands should be run on Pulp 3 machine.
...
+    orphan cleanup task in Pulp 3.
+
+There are cases, when one needs to run a migration from scratch (by default, it's always
+incremental). E.g. some issue happened during the pulp2to3 migration which you are not able to
+recover from.
+
+All the commands should be run on Pulp 3 machine.
+
+
+1. Create a :doc:`Migration Plan <../migration_plan>` which you want to run a migration for.
```
dralley, line 89: s/which you want to run a migration for/for the plugins or set of repositories you would like to migrate/

pulp_2to3_migration/app/tasks/reset.py
```diff
@@ -0,0 +1,81 @@
...
+                # after content is removed we can delete the remaining repositories
+                repo_model.objects.all().only('pk').delete()
+
+            pb.increment()
+
+    pb_data = dict(message="Removing pre-migrated data", code="reset.premigrated.data", total=1)
+    with ProgressReport(**pb_data) as pb:
+        for plugin in plan.get_plugin_plans():
+            for content_type, pulp2to3_content_model in plugin.migrator.content_models.items():
+                pulp2to3_content_model.objects.all().only('pk').delete()
```
dralley, line 66: Does ```.only()``` have an effect even when you're deleting?  That's interesting, I would have hoped Django was a little more clever.

docs/workflows.rst
dralley: But we should probably mention that any Pulp 3 repositories / remotes / everything that were created without being migrated will be deleted as well, for the specified plugins.

docs/workflows.rst
```diff
@@ -64,3 +67,59 @@ All the commands should be run on Pulp 3 machine.
...
+
+.. code:: bash
+
+    $ # migrate content for Pulp 2 ISO plugin
```
dralley, line 93: Leftover from copy-paste

docs/workflows.rst
```diff
@@ -64,3 +67,59 @@ All the commands should be run on Pulp 3 machine.
...
+    $ http POST :24817/pulp/api/v3/migration-plans/ plan='{"plugins": [{"type": "iso"}]}'
+
+    HTTP/1.1 201 Created
+    {
+        "pulp_created": "2019-07-23T08:18:12.927007Z",
+        "pulp_href": "/pulp/api/v3/migration-plans/59f8a786-c7d7-4e2b-ad07-701479d403c5/",
+        "plan": "{ "plugins": [{"type": "iso"}]}"
+    }
+
+2. Use the ``pulp_href`` of the created Migration Plan to reset Pulp 3 data
```
dralley, line 103: Maybe be a little more clear about the special endpoint, it's only shown in the script and not mentioned in the instruction.

dralley: That applies to the ```run/``` endpoint this was copied from as well, I think.

dralley:
Apart from documentation suggestions this looks great, thanks @goosemania. 

It would be great if we could test this, but a functional test is difficult without access to the models, and a unit test is difficult due to the tasking system.  No good answers there.

pulp_2to3_migration/app/tasks/reset.py
goosemania: I was getting OOM very quickly on my small reproducer setup without it.
I can retest to confirm. I believe packages are the main source of the problem because of the metadata stored in the premigrated tables.

docs/workflows.rst
```diff
@@ -64,3 +68,62 @@ All the commands should be run on Pulp 3 machine.
             }
         ]
     }
+
```
dralley, line 71: Typo "migraiton" on line 43

dralley (approved):

dralley (approved):

PR #265, opened 2020-11-11
Small cleanup, remove unneeded query

dralley:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -587,10 +587,6 @@ def pre_migrate_repocontent(repo):
 
     mongo_repocontent_q = mongo_Q(repo_id=repo.pulp2_repo_id)
     mongo_repocontent_qs = RepositoryContentUnit.objects(mongo_repocontent_q)
-    if not mongo_repocontent_qs:
-        # Either the repo no longer exists in Pulp 2,
-        # or the repo is empty.
-        return
 
     repocontent = []
     for repocontent_data in mongo_repocontent_qs.exclude('repo_id').as_pymongo().no_cache():
@@ -602,6 +598,10 @@ def pre_migrate_repocontent(repo):
             pulp2_updated=repocontent_data['updated']
         )
         repocontent.append(item)
+
+    if not repocontent:
+        # Either the repo no longer exists in Pulp 2, or the repo is empty.
+        return
 
     Pulp2RepoContent.objects.bulk_create(repocontent, batch_size=DEFAULT_BATCH_SIZE)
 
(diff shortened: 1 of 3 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -578,19 +578,12 @@ def pre_migrate_repocontent(repo):
     Args:
         repo(Pulp2Repository): A pre-migrated pulp 2 repository which importer should be migrated
     """
-    if repo.is_migrated:
-        return
```
dralley, line 582: This function is called from only one place, and in that place, ```is_migrated``` will always be false.

https://github.com/pulp/pulp-2to3-migration/blob/e45737ad9c82978242e760be4bd5354994b2a2c7/pulp_2to3_migration/app/pre_migration.py#L405-L428

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -578,19 +578,12 @@ def pre_migrate_repocontent(repo):
...
-
     # At this stage the pre-migrated repo is either new or changed since the last run.
     # For the case when something changed, old repo-content relations should be removed.
     Pulp2RepoContent.objects.filter(pulp2_repository=repo).delete()
 
     mongo_repocontent_q = mongo_Q(repo_id=repo.pulp2_repo_id)
     mongo_repocontent_qs = RepositoryContentUnit.objects(mongo_repocontent_q)
-    if not mongo_repocontent_qs:
```
dralley, line 590: Here's the query evaluation.

pulp_2to3_migration/app/pre_migration.py
goosemania: I agree. My concern that it's very implicit and if we change logic at any time in the piece of code you are referring to, we might miss it.
Even though the code is redundant, it's explicit and we are playing it safe.
What do you think?

dralley: Well, if we change the logic in other ways, the defensive checks could cause their own bugs :)

I kind of see your point, but I'd rather do an assertion than try to silently do nothing if we ever use the function wrong.

goosemania: What kind of an assertion do you have in mind?

I don't think there is any correctness concern, just a performance one.
If we run `re_migrate_repocontent` for a migrated repo, it's more work and the task takes longer. I think that's it.

In any case, I don't have a strong opinion here.

goosemania (approved):

pulp_2to3_migration/app/pre_migration.py
dralley: My point is just that we're currently doing this correctly, and if we ever break it in the future such that we do accidentally call ```migrate_repocontent()``` on migrated repos, I'd rather it fail loudly just in case we made the same mistake with anything else.

goosemania: In theory and as a general approach, I agree.
Here we won't fail loudly, we'll just silently redo some work which we could have avoided otherwise.
So pure performance consequences if any.

dralley: Well, I don't feel that strongly about it either, so if you prefer it the way it was before I can revert.  In any case we hopefully won't break this in the first place.

PR #266, opened 2020-11-13
Mark importers and distributors as not_in_plan properly

goosemania:
closes #7823
https://pulp.plan.io/issues/7823

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -617,50 +617,49 @@ def handle_outdated_resources(plan, type_to_repo_ids):
         type_to_repo_ids(dict): A mapping from a pulp 2 repo type to a list of pulp 2 repo_ids
     """
     for plugin_plan in plan.get_plugin_plans():
-        repos = plugin_plan.get_repositories()
+        inplan_repos = plugin_plan.get_repositories()
 
         # filter by repo type
         repos_to_consider = type_to_repo_ids[plugin_plan.type]
 
         # in case only certain repositories are specified in the migration plan
-        if repos:
-            repos_to_consider = set(repos).intersection(repos_to_consider)
+        if inplan_repos:
+            repos_to_consider = set(inplan_repos).intersection(repos_to_consider)
 
         mongo_repo_q = mongo_Q(repo_id__in=repos_to_consider)
+        mongo_repo_obj_ids = set(str(i.id) for i in Repository.objects(mongo_repo_q).only('id'))
 
-        mongo_repo_object_ids = set(
-            str(i.id) for i in Repository.objects(mongo_repo_q).only('id'))
+        repo_type_q = Q(pulp2_repo_type=plugin_plan.type)
+        inplan_repo_q = Q(pulp2_object_id__in=mongo_repo_obj_ids)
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -617,50 +617,49 @@ def handle_outdated_resources(plan, type_to_repo_ids):
...
+            repos_to_consider = set(inplan_repos).intersection(repos_to_consider)
 
         mongo_repo_q = mongo_Q(repo_id__in=repos_to_consider)
+        mongo_repo_obj_ids = set(str(i.id) for i in Repository.objects(mongo_repo_q).only('id'))
 
-        mongo_repo_object_ids = set(
-            str(i.id) for i in Repository.objects(mongo_repo_q).only('id'))
+        repo_type_q = Q(pulp2_repo_type=plugin_plan.type)
+        inplan_repo_q = Q(pulp2_object_id__in=mongo_repo_obj_ids)
+        Pulp2Repository.objects.filter(repo_type_q).exclude(inplan_repo_q).update(not_in_plan=True)
```
goosemania, line 634: Logic for the repos  didn't change, I just simplified queries and made it in the same way as importers and distributors.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -617,50 +617,49 @@ def handle_outdated_resources(plan, type_to_repo_ids):
         type_to_repo_ids(dict): A mapping from a pulp 2 repo type to a list of pulp 2 repo_ids
     """
     for plugin_plan in plan.get_plugin_plans():
-        repos = plugin_plan.get_repositories()
+        inplan_repos = plugin_plan.get_repositories()
```
ipanova, line 620: :+1:

ipanova (approved):

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -617,50 +617,49 @@ def handle_outdated_resources(plan, type_to_repo_ids):
...
-        premigrated_imps = Pulp2Importer.objects.filter(pulp2_type_id__in=imp_types)
-        premigrated_imp_object_ids = set(
-            premigrated_imps.values_list('pulp2_object_id', flat=True))
-        removed_imp_object_ids = premigrated_imp_object_ids - mongo_imp_object_ids
 
-        Pulp2Importer.objects.filter(
-            pulp2_object_id__in=removed_imp_object_ids).update(not_in_plan=True)
+        imp_type_q = Q(pulp2_type_id__in=imp_types)
+        inplan_imp_q = Q(pulp2_object_id__in=mongo_imp_obj_ids)
+        Pulp2Importer.objects.filter(imp_type_q).exclude(inplan_imp_q).update(not_in_plan=True)
```
ipanova, line 648: case when an importer/distributor won't be present in pulp2 anymore but will still be specified in the plan will produce `not_in_plan=True` which is a bit misleading but i don't think we can do much about this unless we have 2 flags that distinguish this situation

goosemania: We discussed it int he past and were not able to come up with the better name.
`dont_care` is an option ;)
`Pulp2Importer.objects.filter(imp_type_q).exclude(inplan_imp_q).update(dont_care=True)`

PR #268, opened 2020-11-19
Add an option to skip corrupted or missing Pulp 2 content.

goosemania:
closes #7538
https://pulp.plan.io/issues/7538

### DIFF
--- a/docs/_static/api.json
+++ b/docs/_static/api.json
@@ -84,27 +84,7 @@
                         "content": {
                             "application/json": {
                                 "schema": {
-                                    "type": "object",
-                                    "properties": {
-                                        "count": {
-                                            "type": "integer",
-                                            "example": 123
-                                        },
-                                        "next": {
-                                            "type": "string",
-                                            "nullable": true
-                                        },
-                                        "previous": {
-                                            "type": "string",
-                                            "nullable": true
-                                        },
-                                        "results": {
-                                            "type": "array",
-                                            "items": {
-                                                "$ref": "#/components/schemas/pulp_2to3_migration.MigrationPlanResponse"
-                                            }
-                                        }
-                                    }
+                                    "$ref": "#/components/schemas/Paginatedpulp_2to3_migration.MigrationPlanResponseList"
                                 }
                             }
                         },
@@ -148,7 +128,7 @@
                     }
                 ],
                 "responses": {
-                    "200": {
+                    "201": {
                         "content": {
                             "application/json": {
                                 "schema": {
@@ -244,6 +224,66 @@
                 "responses": {
                     "204": {
                         "description": "No response body"
+                    }
+                }
+            }
+        },
+        "{pulp_2to3_migration_migration_plan_href}reset/": {
+            "post": {
+                "operationId": "migration_plans_reset",
+                "description": "Trigger an asynchronous task to remove data from Pulp 3 related to the plugins specified in the migration plan.",
+                "summary": "Reset Pulp 3 data for plugins specified in the migration plan",
+                "parameters": [
+                    {
+                        "in": "path",
+                        "name": "pulp_2to3_migration_migration_plan_href",
+                        "schema": {
+                            "type": "string"
+                        },
+                        "required": true
+                    }
+                ],
+                "tags": [
+                    "Migration-Plans"
+                ],
+                "requestBody": {
+                    "content": {
+                        "application/json": {
+                            "schema": {
+                                "$ref": "#/components/schemas/pulp_2to3_migration.MigrationPlan"
+                            }
+                        },
+                        "application/x-www-form-urlencoded": {
+                            "schema": {
+                                "$ref": "#/components/schemas/pulp_2to3_migration.MigrationPlan"
+                            }
+                        },
+                        "multipart/form-data": {
+                            "schema": {
+                                "$ref": "#/components/schemas/pulp_2to3_migration.MigrationPlan"
+                            }
+                        }
+                    },
+                    "required": true
+                },
+                "security": [
+                    {
+                        "cookieAuth": []
+                    },
+                    {
+                        "basicAuth": []
+                    }
+                ],
+                "responses": {
+                    "202": {
+                        "content": {
+                            "application/json": {
+                                "schema": {
+                                    "$ref": "#/components/schemas/AsyncOperationResponse"
+                                }
+                            }
+                        },
+                        "description": ""
                     }
                 }
             }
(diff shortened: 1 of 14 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -120,9 +122,16 @@ async def create_artifact(self, pulp2_storage_path, expected_digests={}, expecte
...
-                                              expected_digests=expected_digests,
-                                              expected_size=expected_size)
+        try:
+            artifact = Artifact.init_and_validate(pulp2_storage_path,
+                                                  expected_digests=expected_digests,
+                                                  expected_size=expected_size)
+        except (DigestValidationError, FileNotFoundError, SizeValidationError):
+            if self.skip_corrupted:
+                _logger.warn(f'The content located in {pulp2_storage_path} is missing or '
+                             f'corrupted. It was skipped during Pulp 2to3 migration.')
```
goosemania, line 132: @jlsherrill , let me know if you have any preference for the phrasing or for the log level.

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -53,11 +53,16 @@ class DebMigrator(Pulp2to3PluginMigrator):
...
         """
         Migrate pre-migrated Pulp 2 Debian content.
+
+        Args:
+            skip_corrupted (bool): If True, corrupted content is skipped during migration,
+                                   no task failure.
+
         """
-        first_stage = ContentMigrationFirstStage(cls)
+        first_stage = ContentMigrationFirstStage(cls, skip_corrupted=skip_corrupted)
```
goosemania, line 65: @quba42 , changes in the debian part, please review.

dralley (approved):
Looks good.  It would still be best however if we didn't run into this issue in the first place.  It looks like the pulp-integrity tool does calculate checksums, maybe we should recommend that it be used prior to a migration to sniff out potential issues? 

https://github.com/dparalen/pulp-integrity/blob/8dfecb0943462d9899d47dcffde443d8779e755e/pulp_integrity/generic.py#L155

goosemania:
> Looks good. It would still be best however if we didn't run into this issue in the first place. It looks like the pulp-integrity tool does calculate checksums, maybe we should recommend that it be used prior to a migration to sniff out potential issues?
> 
> https://github.com/dparalen/pulp-integrity/blob/8dfecb0943462d9899d47dcffde443d8779e755e/pulp_integrity/generic.py#L155

The problem is that potentially not all content can be fixed (e.g. remote url is not responding anymore) and Katello doesn't provide a way to remove content, so users can't fix content if it's broken-broken. And the thought is that if they didn't notice or didn't care about it in pulp 2, it's likely that they don't need it in pulp 3 and we just skip it.
It's a good point and worth documenting in katello that we encourage to do a sanity check upfront but users are users and not everyone will do it.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -120,9 +122,16 @@ async def create_artifact(self, pulp2_storage_path, expected_digests={}, expecte
...
+                return
+            raise
```
ipanova, line 134: @goosemania  Do you think it will be valuable if we would raise a more migration friendly error something along the lines " the artifact is whether missing or is corrupted please fix it before proceeding or use 'skip corrupted content option in the next run'

goosemania: yeah, can do.

ipanova: did we test the scenario when a missing file has been re-added and then migration re-run?

goosemania: added in pulp2?

ipanova: yes in pulp2

docs/workflows.rst
```diff
@@ -28,7 +28,8 @@ All the commands should be run on Pulp 3 machine.
 
 
 2. Use the ``pulp_href`` of the created Migration Plan and its ``run/`` endpoint to run the
-migration.
+migration. Optionally, skip corrupted or missing Pulp 2 content by specifying the
+``skip_corrupted=True`` parameter.
```
ipanova, line 32: what about mentioning that  "with the next sync the missing content should be brought back" or remind user to run repo sync if skip corrupted content option has been used"

pulp_2to3_migration/app/plugin/content.py
goosemania: Nope, I haven't tested that.
If the file is the only thing that changed (no changes in the timestamp), it won't be remigrated on the next run.
We need to document that!

docs/workflows.rst
goosemania: I didn't want to overload the worflow section.
Maybe we need a page which will document all options in greater detail than it's done here.

pulp_2to3_migration/app/plugin/content.py
ipanova: The reason i have asked question about the missing file is to figure out how useful 'skip corrupted' option is - if the user still cannot migrate the data even if he has fixed the missing file then maybe it would be better to just issue the warning and not raise the error.

goosemania: Tested. All good.

docs/workflows.rst
goosemania: Added a more elaborate description to the REST API docs.

docs/_static/api.json
```diff
@@ -770,13 +770,100 @@
...
                         "type": "boolean",
                         "writeOnly": true,
                         "default": false,
                         "description": "If ``True``, performs validation of a Migration Plan only, no migration is run."
+                    },
+                    "skip_corrupted": {
+                        "type": "boolean",
+                        "writeOnly": true,
+                        "default": false,
+                        "description": "If ``True``, skips corrupted or missing Pulp 2 content without causing a task failure. If you need this content, run a sync task in Pulp 3 for a repo of interest to bring any missing content back. Alternatively, repair it in Pulp 2and re-run the migration task."
```
goosemania, line 785: added more here

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -120,9 +123,19 @@ async def create_artifact(self, pulp2_storage_path, expected_digests={}, expecte
...
+        try:
+            artifact = Artifact.init_and_validate(pulp2_storage_path,
+                                                  expected_digests=expected_digests,
+                                                  expected_size=expected_size)
+        except (DigestValidationError, FileNotFoundError, SizeValidationError):
+            if self.skip_corrupted:
+                _logger.warn(f'The content located in {pulp2_storage_path} is missing or '
+                             f'corrupted. It was skipped during Pulp 2to3 migration.')
+                return
+            raise ArtifactValidationError(f'The content located in {pulp2_storage_path} is '
```
ipanova, line 135: :+1:

ipanova (approved):
thank you!

PR #269, opened 2020-11-20
Add Github Actions CI

dralley:
[noissue]


goosemania:
Are the custom scripts going to be used? E.g. https://github.com/pulp/pulp-2to3-migration/blob/master/.travis/post_before_install.sh and https://github.com/pulp/pulp-2to3-migration/blob/master/.travis/post_before_script.sh

It's needed because we have somewhat special setup with mongo parts.

dralley:
@goosemania I just need to copy those scripts.  Will-do.

dralley:
@goosemania done. once this is merged I'll make test PR and fix any issues still remaining.

goosemania (approved):

PR #270, opened 2020-11-20
Make CI adjustments

dralley:
[noissue]


ipanova (approved):

goosemania:
I believe the failure is caused by the fact that pulp_deb is not installed.
It is configured in the template_config.yaml. But scripts don't include it for some reason. Is it possible that some outdated scripts are used?

PR #271, opened 2020-11-23
Fix the project check in GHA

goosemania:
[noissue]


dralley (approved):
@goosemania Thanks, it looks like all of these snuck their way into the template itself, if you would like to open a matching PR over there.

ipanova (approved):

PR #273, opened 2020-11-24
Migrate Debian content types for structured publishing

quba42:
fixes #7865
https://pulp.plan.io/issues/7865

Requires: https://github.com/pulp/pulp-2to3-migration/pull/295


pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -184,3 +184,104 @@ def create_pulp3_content(self):
...
+        )
+
+    def create_pulp3_content(self):
+        """
+        Create a Pulp 3 detail Content unit for saving it later in a bulk operation.
+
+        Return an unsaved Pulp 3 Content
+        """
+        release_relation = {'release': '???'}
+        return (pulp3_models.ReleaseComponent(component=self.component), release_relation)
```
quba42, line 287: I opened this Draft PR, because I need some help:
The Pulp3 `ReleaseComponent` I am creating here, has the field `release` of type `ForeignKey` that should point at a Pulp3 content unit of type `Release`. I suspect that this relation should be returned to the calling instance via what I have here called `release_relation`. However, I have no clear understanding of what `release_relation` needs to contain for the mechanism to function as designed. I want a reference to the `Release` component unit, which will end up in the same repository version as the `ReleaseComponent` I am creating, and also has the same value in it's `distribution` field, as `self.distribution`. Any help is appreciated!

quba42: Extra Information: If I run the code as is (with `release_relation = {'release': '???'}`) I get the following error:
```
pulp_deb.app.models.content.ReleaseComponent.DoesNotExist: ReleaseComponent matching query does not exist.
```

quba42: Ok, I found something that will allow me to continue for now. The docker migration pipeline has a special stage that picks up on the things that are returned using what I have called `release_relation` and processes them. I will try to emulate that mechanism.

goosemania: @quba42 , you should try to reuse the stages from your sync pipeline.
If you have this relation, I guess the sync pipeline has a stage or some code that deals with it, creates this relation.

quba42: In the Debian sync pipeline, this is done right in the first stage, by ensuring one of the two related partners is pushed all the way through the pipeline right away, and is then immediately available for relating it to the other content type. That wont work for the migration pipeline, since the migration pipeline loops over each content type in turn and does not have both of them available at the same time at the start. However, I think I can emulate what docker is doing and retrieve the "to be associated" unit in a later dedicated `InterrelateContent` stage.

Will take me a while to implement though, especially since other priorities are intruding on my time again...

quba42:
Migrating releases and components now works, but I still need to add `ReleaseArchitecture` and `PackageReleaseComponent` creation.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -184,3 +184,257 @@ def create_pulp3_content(self):
...
+        Create a Pulp 3 detail Content unit for saving it later in a bulk operation.
+
+        Return an unsaved Pulp 3 Content
+        """
+        release_component = pulp3_models.ReleaseComponent.objects.filter(
+            component=self.component,
+            release__distribution=self.distribution,
+            release__codename=self.codename,
+            release__suite=self.suite,
+        ).first()
```
quba42, line 388: In the current implementation retrieving this `release_component` (for associating it) appears to fail, because it was never saved to the data base. However, if I comment out or remove lines 50 and 51 the pipeline does finish successfully. Therefore I know that creating release components does work in principle, and I also know that retrieving units for association does work in principle (since creating release components also does this).

I can only conclude that this is some order issue, where python is trying to retrieve the release_component before it is saved to the DB. However, I was under the impression that content types are migrated in turn, and thus should guarantee order. (Otherwise what is the point of using an ordered dict in line 46 and the comment found [here](https://github.com/pulp/pulp-2to3-migration/blob/ed3fdc3457dce5fc1981acd26f779ef16a587944/pulp_2to3_migration/app/plugin/content.py#L178)?)

I have been stuck on this for a while now, and any help would be greatly appreciated.
It might be easier to explain in a screen session, so I would always be up for one of those.

goosemania: I'm struggling with one thing, please let me know where my understanding/logic is off.
`pre_migrate_content_detail` does nothing, it means that all values for this class (Pulp2DebComponentPackage) are null or empty string, correct? If so, then do you expect a ReleaseComponent with all empty values, e.g. self.distribution, self.component, etc?

And what are the lines 50/51 do you refer to? in which module? Thanks!

quba42: Wups, I am referring to lines 50, 51 in `pulp_2to3_migration/app/plugin/deb/migrator.py`.

The fact that `pre_migrated_content_detail` is empty here is a bit of a dirty workaround around the plugin design.
I basically need to create several pulp 3 types from a single pulp 2 type. To avoid having to retrieve the pulp2 units from mongo multiple times, I made the `pre_migrate_content_detail` from `Pulp2DebComponent` create not only `Pulp2DebComponent` objects, but also `Pulp2DebComponentPackage` and `Pulp2DebReleaseArchitecture` objects.
The `pre_migrate_content_detail` functions from the latter two classes then no longer need to do anything. Nevertheless all the pre migrated content I need is created correctly.

quba42: Now that I read that, I guess that does not actually avoid batches from being retrieved from mongo and passed to the functions that then don't do anything... Maybe I should just split up the `pre_migrated_content_detail` function from `Pulp2DebComponent` into three.

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -35,13 +37,19 @@ class DebMigrator(Pulp2to3PluginMigrator):
...
     pulp3_repository = pulp3_models.AptRepository
-    content_models = {
-        'deb': pulp_2to3_models.Pulp2DebPackage,
-    }
+    content_models = OrderedDict([
+        ('deb_release', pulp_2to3_models.Pulp2DebRelease),
+        ('deb', pulp_2to3_models.Pulp2DebPackage),
+        ('deb_component', pulp_2to3_models.Pulp2DebComponent),
+        ('deb_component2', pulp_2to3_models.Pulp2DebComponentPackage),
+        ('deb_component3', pulp_2to3_models.Pulp2DebReleaseArchitecture),
```
quba42, line 51: Perhaps I need to explain a bit more what I did here:

As far as I can tell the migration plugin design more or less assumes that exactly one pulp 2 type will be mapped on exactly one pulp 2 type. However, for `pulp_deb` there are two new pulp 3 types (`ComponentPackage`, `ReleaseArchitecture`) that don't really have any equivalent pulp 2 type. The information to create these Pulp 3 types as well as (`Component`) can be extracted from the single pulp 2 `DebComponent` type (more or less).

As a result I needed several 2to3 types for only a single pulp 2 type. I am fairly confident that the keys in this dict don't actually need to be the name of the corresponding pulp 2 type so I just named them `deb_component`,  `deb_component2` and `deb_component3` here, but the entire thing is a bit workaroundish...

pulp_2to3_migration/app/plugin/deb/migrator.py
```diff
@@ -51,6 +59,12 @@ class DebMigrator(Pulp2to3PluginMigrator):
     future_types = {
         'deb': pulp_2to3_models.Pulp2DebPackage,
```
quba42, line 60: I found that adding my new metadata types here changes things. (I no longer run into an error, but I don't get exactly the result I want either.)

This gives me something I can continue investigating for now.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
quba42: I have now updated the docstrings to make clear why some `pre_migrate_content_detail` methods don't do anything.
My original issue was fixed by adding the relevant types to the `future_types` list in the migrator.

Resolved.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -184,3 +190,307 @@ def create_pulp3_content(self):
...
+                component=component,
+                suite=suite,
+                pulp2content=pulp2_base_record,
+            ))
+            architectures = set()
+            for package_id in component_unit.packages:
+                package_unit = pulp2_models.DebPackage.objects.filter(id__in=[package_id]).first()
+                package_relative_path = package_unit.filename
+                package_sha256 = package_unit.checksum
+                # TODO: Consider hashing the subid string to ensure it will fit in the pulp2_subid
```
quba42, line 312: I need a second opinion on this. Does this make sense?

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -184,3 +190,307 @@ def create_pulp3_content(self):
...
+                pulp2_subid = (suite + codename + distribution + component
+                               + package_relative_path + package_sha256)
+
+                pulp2_sub_record = Pulp2Content(
```
quba42, line 316: @goosemania I ended up not adding any `get_pulp2_subids` function and instead simply creating the "sub records" as part of the `pre_migrate_content_detail` function.

The reason for this, is that I only know how many "sub records" I will need (what there subids should be) once I run `pre_migrate_content_detail`. I am pretty confident it should not be a problem that they are created slightly later in the migration process than the other `Pulp2Content` units...

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
quba42, line 312: I need a second opinion on this. Does this make sense?

quba42 (commented):
I am thinking it might be a good idea to turn the bottom commit into a separate issue and PR (and merging this PR afterwards). (I would squish all the remaining commits into one before merging).

They strike me as two separate issues and it might be good to have separate release note entries and keep the two migrations separate as well:

- Issue 1: Add pulp2_subid to Pulp2Content model for the sake of one to many 2to3 mappings
- Issue 2: Add ability to migrate structured Debian content

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
quba42: I am basically ensuring that my `pulp2_subid`'s are unique by concatenating all the fields from my "unique together" conditions into a single string and using that. I am a bit worried that the concatenated strings might exceed the `max_length=255` property of the `pulp2_subid` field.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
goosemania: I'm not sure how it works...
[Pulp2Content is already saved by this time](https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/pre_migration.py#L211-L217).
I haven't tested your PR, however I read it as you create new Pulp2Content which now have subids and which  you link to the detail model.
So what happens with the Pulp2Content created before that which would have subid = null?

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
goosemania: +1 on hashing it

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
quba42: Every Pulp 2 `DebComponent` from Mongo is ultimately turned into exactly one Pulp 3 `ReleaseComponent`, as well as `n` `PackageReleaseComponent` and `m` `ReleaseArchitecture` entries.

The initial `Pulp2Content` entry (with `pulp2_subid = ''`) is associated with the (exactly one) Pulp 3 `ReleaseComponent`. I call this initial `Pulp2Content` entry the `pulp2_base_record` in the code. So far this is exactly the same as for every other type.

In addition, I am having `Pulp2DebComponent.pre_migrate_content_detail` create additional `Pulp2Content` units for each `Pulp2DebComponentPackage` and each `Pulp2DebReleaseArchitecture` I create. These additional `Pulp2Content` entries are given the same values as the corresponding `pulp2_base_record` for every field except the `pulp2_subid` field. The `pulp2_subid` field is generated from the `unique_together` fields of the relevant `Pulp2DebComponentPackage`/`Pulp2DebReleaseArchitecture`, which have a foreign key field on the newly created `Pulp2Content` `pulp2_sub_record`.

That is really all there is to it. Everything else appears to work as expected out of the box. There is now exactly one  `Pulp2Content` record for every Pulp3 entry that is eventually created and they are already correctly associated. The migration will mark them as migrated whenever the relevant Pulp3 content is created just like for everything else.

quba42: pulp_2to3_migration_flow
Not sure if this helps...

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
quba42: Done. I also got rid of my remaining `TODO` comments.

quba42:
Note: I have retested this latest state manually. It still migrates my test repository as expected. My full test cases was something like the following:

1. Obtaining the test repo using pulp2 (for migration):
```
pulp-admin login -u admin -p admin
pulp-admin deb repo create --repo-id 'tuxedo_repo' --feed 'https://deb.tuxedocomputers.com/ubuntu/' --releases 'bionic' --components 'main' --architectures 'i386,amd64' --relative-url 'tuxedo'
pulp-admin deb repo sync run --repo-id='tuxedo_repo'
```

2. Add and execute the migration plan:
```
BASE_ADDR=':'
MIGRATION_PLAN_HREF=$(http post "${BASE_ADDR}/pulp/api/v3/migration-plans/" plan='{"plugins": [{"type": "deb"}]}' |  jq -r '.pulp_href')
MIGRATION_TASK_HREF=$(http post "${BASE_ADDR}${MIGRATION_PLAN_HREF}run/" | jq -r '.task')
http get "${BASE_ADDR}${MIGRATION_TASK_HREF}" | jq '.state'
```

3. Manually check the resulting Pulp 3 publications, distributions, repositories, content, etc.

pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
goosemania: Thanks for the detailed explanation!

goosemania (approved):

PR #274, opened 2020-11-27
Add ability to apply a pulp 2 snapshot from within a test

goosemania:
closes #7903
https://pulp.plan.io/issues/7903

### DIFF
--- a/pulp_2to3_migration/tests/functional/util.py
+++ b/pulp_2to3_migration/tests/functional/util.py
@@ -1,6 +1,16 @@
+import logging
+import os
+import subprocess
+
 from time import sleep
 
 from .dynaconf_config import settings
+
+from pulp_smash import cli, utils
+from pulp_smash import config as smash_config
+
+
+_logger = logging.getLogger(__name__)
 
 
 def monitor_task(tasks_api, task_href):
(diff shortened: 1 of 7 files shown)


pulp_2to3_migration/tests/functional/util.py
```diff
@@ -48,3 +58,33 @@ def get_psql_smash_cmd(sql_statement):
...
+    Args:
+        name(str): directory name  of the pulp 2 snapshot
+
+    """
+    smash_cfg = smash_config.get_config()
+    smash_cli_client = cli.Client(smash_cfg)
+    pulp2_mongodb_conf = utils.get_pulp_setting(smash_cli_client, "PULP2_MONGODB")
+    mongodb_host = pulp2_mongodb_conf['seeds'].split(':')[0]
+
+    pulp2_fs_setup_script_path = '/tmp/set_pulp2.sh'
```
dralley, line 76: Is this file supposed to be created externally?

goosemania: This script needs to be on the machine where pulp runs. Doesn't matter where. /tmp/ was an easy location to pick, any suggestions are welcome.
I saw pulp_deb also copying their scripts into /tmp/, probably for the same reason.

dralley: It's something that a test is expected to do themselves, right?

I just wanted to make sure none of the code here is supposed to be creating it.

goosemania: It is copied to the container by the CI script https://github.com/pulp/pulp-2to3-migration/pull/274/files#diff-de486dbe208dad5a436ac7b007fe0db87114bc32af1aa5693bd088ff52cac8f8R19. So the code in this specific module doesn't care about putting it in place, it just using it.

goosemania: A test it self is supposed to call  this `set_pulp2_snaphot` util. Not necessarily each test but the setup class/steps.

dralley: gotcha

dralley: Well, I don't have any ideas on how to do this more cleanly, and I'm kind of resigned to the fact that the migration infra is going to be a mess no matter how much time we spend on it, so I have no objection to merging.

goosemania:
@dkliban , @dralley , @ipanova , please review, anyone, so we could merge it and build tests on top of it.

dkliban (approved):
This looks good to me. Thanks for putting it together!

ipanova (approved):

PR #279, opened 2020-12-08
Added migraiton plan tests

goosemania:
Also switched to using the renamed pulp2 snapshot.

closes #7934
https://pulp.plan.io/issues/7934

### DIFF
--- a/pulp_2to3_migration/tests/functional/test_file_migration.py
+++ b/pulp_2to3_migration/tests/functional/test_file_migration.py
@@ -1,4 +1,5 @@
 import json
+import time
 import unittest
 
 from pulpcore.client.pulpcore import (
@@ -124,7 +125,7 @@ class TestMigrationPlan(unittest.TestCase):
         cls.tasks_api = TasksApi(core_client)
         cls.migration_plans_api = MigrationPlansApi(migration_client)
 
-        set_pulp2_snapshot(name='20191031')
+        set_pulp2_snapshot(name='file_base_4repos')
 
     def tearDown(self):
         """
@@ -132,6 +133,7 @@ class TestMigrationPlan(unittest.TestCase):
         """
         cmd = get_psql_smash_cmd(TRUNCATE_TABLES_QUERY_BASH)
         self.smash_cli_client.run(cmd, sudo=True)
+        time.sleep(0.5)
 
     def _do_test(self, repos, migration_plan):
         mp = self.migration_plans_api.create({'plan': migration_plan})
--- a/pulp_2to3_migration/tests/functional/test_migration_plan.py
+++ b/pulp_2to3_migration/tests/functional/test_migration_plan.py
@@ -1,0 +1,169 @@
+import json
+import unittest
+
+from pulpcore.client.pulpcore import (
+    ApiClient as CoreApiClient,
+    Configuration,
+    TasksApi
+)
+from pulpcore.client.pulp_file import (
+    ApiClient as FileApiClient,
+    ContentFilesApi,
+    RepositoriesFileApi,
+    RepositoriesFileVersionsApi
+)
+from pulpcore.client.pulp_2to3_migration import (
+    ApiClient as MigrationApiClient,
+    MigrationPlansApi
+)
+from pulpcore.client.pulp_2to3_migration.exceptions import ApiException
+
+from pulp_2to3_migration.tests.functional.util import get_psql_smash_cmd, set_pulp2_snapshot
+
+from pulp_smash import cli
+from pulp_smash import config as smash_config
+from pulp_smash.pulp3.bindings import monitor_task, monitor_task_group, PulpTaskError
+
+from .common_plans import FILE_SIMPLE_PLAN, FILE_COMPLEX_PLAN
+from .constants import TRUNCATE_TABLES_QUERY_BASH
+
+
+EXTRA_COMMA_PLAN = '{"plugins": [{"type": "iso"},]}'
+MISSING_RESOURCE_PLAN = json.dumps({
+    "plugins": [{
+        "type": "iso",
+        "repositories": [
+            {
+                "name": "file",
+                "pulp2_importer_repository_id": "non-existing importer",  # policy: immediate
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "non-existing repo",  # content count: iso - 3
+                        "pulp2_distributor_repository_ids": ["non-existing distributor"]
+                    }
+                ]
+            }
+        ]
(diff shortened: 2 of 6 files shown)


pulp_2to3_migration/tests/functional/test_file_migration.py
```diff
@@ -124,14 +125,15 @@ def setUpClass(cls):
...
-        set_pulp2_snapshot(name='20191031')
+        set_pulp2_snapshot(name='file_base_4repos')
 
     def tearDown(self):
         """
         Clean up the database after each test
         """
         cmd = get_psql_smash_cmd(TRUNCATE_TABLES_QUERY_BASH)
         self.smash_cli_client.run(cmd, sudo=True)
+        time.sleep(0.5)
```
goosemania, line 136: Any better ideas how to wait for the truncate operation to finish?
Without this sleep, I was getting `deadlock detected` in postgres every now and then.

pulp_2to3_migration/tests/functional/plans/base.py
```diff
@@ -0,0 +1,3 @@
+from collections import namedtuple
+
+TestSetup = namedtuple('TestSetup', ['plan', 'outcome'])
```
dralley, line 3: I'm not sure I understand the added value of this extra layer of indirection.  Why not just feed these arguments directly into ```_load_and_run()```?

Personally I would probably not use ```mp_validation.py``` either, and just leave them defined in ```test_file_migration.py```

pulp_2to3_migration/tests/functional/test_migration_plan.py
```diff
@@ -0,0 +1,150 @@
...
+        """Test that only one plan can be run at a time"""
+        plan, outcome = plans.ONLY_ONE_PLAN_TEST
+        mp = self.migration_plans_api.create({'plan': plan})
+
+        # run twice
+        mp_run_response = self.migration_plans_api.run(mp.pulp_href, {})
+        with self.assertRaises(ApiException):
+            self.migration_plans_api.run(mp.pulp_href, {})
+
+        # we should do more specific checks but for now I get empty exc for this call :/
```
dralley, line 133: add a TODO so it's more searchable/spottable

pulp_2to3_migration/tests/functional/plans/base.py
goosemania: The attempt was to make it more readable and define plan and expected result in one place, without coming back and forth looking at the plan and figuring out what the result should be for each test.

It might not be the best way. I can put everything into one file all together. However I'd like to reuse some plans, so I thought a place for constants would be good. Throwing all plans into one file looks too much to me, so I transitioned to grouping plans in some way, in some logical(?) way. Currently the logic is "for MP validation tests, put all plans in a dedicated file, for the other set of tests, put the plans in the other file".

Is putting plans in the same file as tests and having the expected results in each test a preferable option? I can do that, no problem.
I'm also open for any third option to approach that.

pulp_2to3_migration/tests/functional/test_migration_plan.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def test_load_simple_plan(self):
+        """Test that a simple Migration Plan can be loaded and run."""
+        task = self._load_and_run(plans.SIMPLE_PLAN_TEST.plan)
+        self.assertEqual(task.state, "completed")
+
+    def test_load_complex_plan(self):
+        """Test that a complex Migration Plan can be loaded and run."""
+        task = self._load_and_run(plans.COMPLEX_PLAN_TEST.plan)
+        self.assertEqual(task.state, "completed")
```
dralley, line 93: This check isn't necessary, ```monitor_task()``` checks the state and raises an exception if it didn't end up as completed.

pulp_2to3_migration/tests/functional/test_migration_plan.py
```diff
@@ -0,0 +1,150 @@
...
+
+    def test_validate_missing_resource(self):
+        """Test that pulp 2 missing resource is noticed and reported."""
+        plan, outcome = plans.MISSING_RESOURCE_PLAN_TEST
+        mp = self.migration_plans_api.create({'plan': plan})
+        mp_run_response = self.migration_plans_api.run(mp.pulp_href, {'validate': True})
+        with self.assertRaises(PulpTaskError) as exc:
+            monitor_task(mp_run_response.task)
+        self.assertEqual(exc.exception.task.state, 'failed')
```
dralley, line 120: Same here - if it didn't fail, exception wouldn't have been raised.

pulp_2to3_migration/tests/functional/test_file_migration.py
dralley: Nope : /

pulp_2to3_migration/tests/functional/plans/base.py
dralley: I'm not strongly against it.  If you plan to use these constants in multiple files then it could make sense -- it just looked like these constants were only meant for the tests in this one file.

dralley: If the tests are being reused (the named tuple things), in what way would they be tested differently each time?

goosemania: tuples are unique, plans are reused.

pulp_2to3_migration/tests/functional/test_migration_plan.py
```diff
@@ -0,0 +1,170 @@
...
+    def test_unknown_key_plan(self):
+        """Test that Migration Plan creation fails if some unknown keys are mentioned in it."""
+        with self.assertRaises(ApiException) as exc:
+            self.migration_plans_api.create({'plan': UNKNOWN_KEY_PLAN})
+            self.assertEqual(exc.code, 400)
+
+    def test_validate_plan(self):
+        """Test that pulp 2 resources are validated."""
+        task = self._load_and_run(FILE_COMPLEX_PLAN, {'validate': True})
+        self.assertEqual(task.state, "completed")
```
dralley, line 136: ^

dralley (approved):
@goosemania This is much better, thank you!

One other small nitpick above otherwise LGTM

pulp_2to3_migration/tests/functional/test_migration_plan.py
goosemania: oh missed it, thanks

goosemania: fixed

PR #282, opened 2020-12-14
Add non-plugin specific migration behaviour tests.

goosemania:
closes #7966
https://pulp.plan.io/issues/7966

### DIFF
--- a/pulp_2to3_migration/tests/functional/test_migration_behaviour.py
+++ b/pulp_2to3_migration/tests/functional/test_migration_behaviour.py
@@ -1,0 +1,347 @@
+import json
+import time
+import unittest
+
+from pulpcore.client.pulpcore import Configuration
+
+from pulpcore.client.pulp_file import (
+    ApiClient as FileApiClient,
+    ContentFilesApi,
+    DistributionsFileApi,
+    PublicationsFileApi,
+    RemotesFileApi,
+    RepositoriesFileApi,
+    RepositoriesFileVersionsApi
+)
+from pulpcore.client.pulp_2to3_migration import (
+    ApiClient as MigrationApiClient,
+    MigrationPlansApi,
+    Pulp2ContentApi,
+    Pulp2RepositoriesApi
+)
+
+from pulp_2to3_migration.tests.functional.util import get_psql_smash_cmd, set_pulp2_snapshot
+
+from pulp_smash import cli
+from pulp_smash import config as smash_config
+from pulp_smash.pulp3.bindings import monitor_task, monitor_task_group
+
+from .common_plans import FILE_SIMPLE_PLAN, FILE_COMPLEX_PLAN
+from .constants import BINDINGS_CONFIGURATION, FILE_MANY_URL, FILE_URL, TRUNCATE_TABLES_QUERY_BASH
+
+
+MANY_INTO_ONE_PLAN = json.dumps({
+    "plugins": [{
+        "type": "iso",
+        "repositories": [
+            {
+                "name": "file",
+                "pulp2_importer_repository_id": "file-many",  # policy: on_demand
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file"  # content count: iso - 3
+                    },
+                    {
+                        "pulp2_repository_id": "file-large"  # content count: iso - 10
+                    },
+                    {
+                        "pulp2_repository_id": "file-many"  # content count: iso - 250
+                    }
+                ]
+            }
+        ]
+    }]
+})
+
+IMPORTER_DIFF_PLAN = json.dumps({
+    "plugins": [{
+        "type": "iso",
+        "repositories": [
+            {
+                "name": "file",
+                "pulp2_importer_repository_id": "file-many",  # policy: on_demand
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file"  # content count: iso - 3
+                    }
+                ]
+            },
+            {
+                "name": "file-many",
+                "pulp2_importer_repository_id": "file",  # policy: immediate
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file-many"  # content count: iso - 250
+                    }
+                ]
+            }
+        ]
+    }]
+})
+
+DISTRIBUTOR_DIFF_PLAN = json.dumps({
+    "plugins": [{
+        "type": "iso",
+        "repositories": [
+            {
+                "name": "file",
+                "pulp2_importer_repository_id": "file",  # policy: immediate
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file",  # content count: iso - 3
+                        "pulp2_distributor_repository_ids": ["file-many"]
+                    }
+                ]
+            },
+            {
+                "name": "file-many",
+                "pulp2_importer_repository_id": "file-many",  # policy: on_demand
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file-many",  # content count: iso - 250
+                        "pulp2_distributor_repository_ids": ["file"]
+                    }
+                ]
+            }
+        ]
+    }]
+})
+
+IMPORTER_NO_REPO_PLAN = json.dumps({
+    "plugins": [{
+        "type": "iso",
+        "repositories": [
+            {
+                "name": "file",
+                "pulp2_importer_repository_id": "file-many",  # policy: on_demand
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file"  # content count: iso - 3
+                    }
+                ]
+            }
+        ]
+    }]
+})
+
+DISTRIBUTOR_NO_REPO_PLAN = json.dumps({
+    "plugins": [{
+        "type": "iso",
+        "repositories": [
+            {
+                "name": "file",
+                "pulp2_importer_repository_id": "file",  # policy: immediate
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file",  # content count: iso - 3
+                        "pulp2_distributor_repository_ids": ["file-many"]
+                    }
+                ]
+            }
+        ]
+    }]
+})
+
+CONTENT_COUNT = {
+    'file': 3,
+    'file2': 3,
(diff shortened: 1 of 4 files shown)


pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -0,0 +1,337 @@
...
+        pulp3_repo = self.file_repo_api.list().results[0]
+        repo_content = self.file_content_api.list(repository_version=pulp3_repo.latest_version_href)
+
+        self.assertEqual(len(self.file_repo_api.list().results), 1)
+        self.assertEqual(self.file_repo_versions_api.list(pulp3_repo.pulp_href).count, 4)
+        for pulp2repo in pulp2repositories:
+            with self.subTest(pulp2repo=pulp2repo):
+                self.assertTrue(pulp2repo.is_migrated)
+                self.assertEqual(pulp2repo.pulp3_repository_href, pulp3_repo.pulp_href)
+        self.assertEqual(repo_content.count, 263)
```
goosemania, line 259: This is currently failing for a reason. I need to investigate it, looks like a bug at  first glance.

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -0,0 +1,337 @@
...
+        Test that an importer can be migrated with any Pulp 2 repo, not its native one.
+
+        Importers are swapped in the plan.
+        """
+        self._load_and_run(IMPORTER_DIFF_PLAN)
+        pulp2repositories = self.pulp2repositories_api.list(ordering='pulp2_id').results
+        pulp2repo1, pulp2repo2 = pulp2repositories
+        pulp3_remote1 = self.file_remotes_api.read(pulp2repo1.pulp3_remote_href)
+        pulp3_remote2 = self.file_remotes_api.read(pulp2repo2.pulp3_remote_href)
+        remote_url_repo1 = 'https://repos.fedorapeople.org/pulp/pulp/fixtures/file-many' \
```
dralley, line 272: I'd suggest either putting the string in parens, or indenting it on the next line so that there is space, or both.

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -0,0 +1,337 @@
...
+        self.assertEqual(pulp3_dist2.base_path, 'file')
+
+    def test_importer_no_repo(self):
+        """Test that an importer can be migrated without its native Pulp 2 repo."""
+        self._load_and_run(IMPORTER_NO_REPO_PLAN)
+        pulp2repository = self.pulp2repositories_api.list().results[0]
+        pulp3_remote = self.file_remotes_api.read(pulp2repository.pulp3_remote_href)
+        remote_url = 'https://repos.fedorapeople.org/pulp/pulp/fixtures/file-many/PULP_MANIFEST'
+
+        self.assertEqual(len(self.pulp2repositories_api.list().results), 1)
```
dralley, line 317: Is there a particular reason to take the len of the list rather than using the count?

dralley: (I mean that the API request returns the count as well as the list of results)

goosemania: no reason, just silly me, will fix it

dralley: It doesn't actually matter, really nitpicky thing

dralley (approved):

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -0,0 +1,337 @@
...
+        file_content = self.file_content_api.read(content_href)
+        self.assertEqual(file_content.relative_path, '137.iso')
+
+    def _test_pulp2repositories(self, plan):
+        """
+        Test that pulp2repositories/ endpoint provides info for the migrated Pulp 2 repository.
+
+        Check correctness of the dara for the first repo in the list.
+        """
+        self._load_and_run(FILE_SIMPLE_PLAN)
```
ipanova, line 210: s/FILE_SIMPLE_PLAN/plan/

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -0,0 +1,337 @@
...
+                "pulp2_importer_repository_id": "file",  # policy: immediate
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "file"  # content count: iso - 3
+                    },
+                    {
+                        "pulp2_repository_id": "file-large"  # content count: iso - 10
+                    },
+                    {
+                        "pulp2_repository_id": "file-many"  # content count: iso - 250
```
ipanova, line 48: file-many is on_demand repo and it's content won't be migrated unless you specify pulp2_importer_id: file-many, that's why line 259 fails since it does not contain 250 units.
However besides this observation it indeed seems that there is an issue in the logic.
New repo_version discards the content from previous repo_version completely, meaning that newly created version does not have 'additive' behaviour https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/migration.py#L369
My expectation is to see additive behaviour, is it correct or i am missing something?

ipanova:
@goosemania tests look good to me, thank you for adding those.
Let's figure out the outstanding failing test :+1:

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
goosemania: Agreed with expectation around the on_demand content.

I had the same thinking process and then thinking more about it decided that additive wouldn't make sense in the migration case. It means no content will be ever removed in the later versions.
 Katello confirmed that they need a 'mirror' behaviour, since pulp repo versions will correspond to CV whichin pulp2 had to be in an individual repo each.

I'll adjust the test to reflect this behaviour.

Thanks for feedback

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
goosemania: The funny thing is that I used count in some places, it's just for those calls it didn't cross my mind :shrug: 
Fixed int he next push ;)

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
goosemania: Did you mean this way?
```
        remote_url_repo1 = (
            'https://repos.fedorapeople.org/pulp/pulp/fixtures/file-many/PULP_MANIFEST')
```

dralley: I would go for

```
remote_url_repo1 = \
            'https://repos.fedorapeople.org/pulp/pulp/fixtures/file-many/PULP_MANIFEST'
```
or

```
remote_url_repo1 = (
            'https://repos.fedorapeople.org/pulp/pulp/fixtures/file-many/PULP_MANIFEST'
)

**Comment by goosemania** [comment]
I believe I addressed all feedback. The test is fixed as well.

File: pulp_2to3_migration/tests/functional/test_migration_behaviour.py [hunk 6]
```diff
@@ -0,0 +1,353 @@
...
+        self.assertEqual(self.file_repo_versions_api.list(pulp3_repo.pulp_href).count, 4)
+        self.assertEqual(self.file_content_api.list().count, 263)
+        for pulp2repo in pulp2repositories:
+            with self.subTest(pulp2repo=pulp2repo):
+                self.assertTrue(pulp2repo.is_migrated)
+                self.assertEqual(pulp2repo.pulp3_repository_href, pulp3_repo.pulp_href)
+                repo_content = self.file_content_api.list(
+                    repository_version=pulp2repo.pulp3_repository_version
+                )
+                self.assertEqual(repo_content.count, CONTENT_COUNT[pulp2repo.pulp2_repo_id])
```
ipanova, line 274: :+1:

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -0,0 +1,353 @@
...
+
+        Importers are swapped in the plan.
+        """
+        self._load_and_run(IMPORTER_DIFF_PLAN)
+        pulp2repositories = self.pulp2repositories_api.list(ordering='pulp2_id').results
+        pulp2repo1, pulp2repo2 = pulp2repositories
+        pulp3_remote1 = self.file_remotes_api.read(pulp2repo1.pulp3_remote_href)
+        pulp3_remote2 = self.file_remotes_api.read(pulp2repo2.pulp3_remote_href)
+        remote_url_repo1 = (
+            'https://repos.fedorapeople.org/pulp/pulp/fixtures/file-many/PULP_MANIFEST'
```
ipanova, line 288: mind updating the links to https://fixtures.pulpproject.org/?

ipanova: since some of the urls are re-used in tests have you considered to make them constants?

goosemania: I do mind at the moment because it means changing the pulp 2 snapshot.
If we make a new pulp2 snaphot we can change it.

I can move them to constants, left it here for readability.

ipanova: you are saying that the snapshot we are working with references these old urls? This is what confused me.

ipanova (approved):

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
ipanova: this is a nictpick but i would name those, for readability FILE_MANY_URL, FILE_ULR, etc.

goosemania: > you are saying that the snapshot we are working with references these old urls? This is what confused me.

yes

PR #291, opened 2021-01-13
Taught content pre_migration to use consistent ordering.

ggainey:
This allows retries after migration-failures to pick up from failure-points consistently.

NOTE: In the presence of large migrations, this code requires the index from
https://github.com/pulp/pulp/pull/4011 to exist in Pulp2, or the migration is likely
to fail at the order-by point.

fixes #7781

### DIFF
--- a/CHANGES/7781.bugfix
+++ b/CHANGES/7781.bugfix
@@ -1,0 +1,5 @@
+Taught pre-migration to order content by last-updated.
+
+This lets a migration recover reliably from fatal errors during migration attempts.
+NOTE: this fix assumes the Pulp2 instance is at least at 2.21.5. Earlier versions are
+missing an index in the Mongo database that makes the ordering possible.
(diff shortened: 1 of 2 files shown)


ipanova (approved):

CHANGES/7781.bugfix
```diff
@@ -0,0 +1,2 @@
+Taught pre-migration to order content by last-updated, in order to be able to reliably
+recover from errors during migration attempts.
```
goosemania, line 2: Do you think it's a bad idea if we also mention here that Pulp 2.21.5+ is required from now on? 
I understand that for ultra small setups it won't be the case. However you mentioned that it fails even with one RHEL repo, correct? So threshold is extremely low :/

ggainey: Fine idea - change incoming.

goosemania (changes requested):
Just so no one merges it before changes are in

CHANGES/7781.bugfix
ggainey: Updated

goosemania (approved):
Thanks!

PR #292, opened 2021-01-14
An attempt to help with pulp_deb issue

goosemania:
where 1 pulp 2 unit becomes N content units of different types and many can be present in one repo.

### DIFF
--- a/pulp_2to3_migration/app/models/content.py
+++ b/pulp_2to3_migration/app/models/content.py
@@ -33,9 +33,10 @@ class Pulp2Content(BaseModel):
     downloaded = models.BooleanField(default=False)
     pulp3_content = models.ForeignKey(Content, on_delete=models.SET_NULL, null=True)
     pulp2_repo = models.ForeignKey(Pulp2Repository, on_delete=models.SET_NULL, null=True)
+    pulp2_subid = models.CharField(max_length=255, null=True)
 
     class Meta:
-        unique_together = ('pulp2_id', 'pulp2_content_type_id', 'pulp2_repo')
+        unique_together = ('pulp2_id', 'pulp2_content_type_id', 'pulp2_repo', 'pulp2_subid')
         indexes = [
             models.Index(fields=['pulp2_content_type_id']),
         ]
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/models/content.py
```diff
@@ -33,9 +33,10 @@ class Pulp2Content(BaseModel):
     downloaded = models.BooleanField(default=False)
     pulp3_content = models.ForeignKey(Content, on_delete=models.SET_NULL, null=True)
     pulp2_repo = models.ForeignKey(Pulp2Repository, on_delete=models.SET_NULL, null=True)
+    pulp2_subid = models.CharField(max_length=255, null=True)
```
goosemania, line 36: There is a use case when one pulp2 unit migrates into multiple content units of different type.
This is not a working PR but more of an idea how to workaround that problem, if we want to keep pulp2content/ endpoint flat and not change pulp3_content to many2many field.

Alternatively we can make it many2many and flaten everything usiung serializer so REST API stays as it currently is.

Any thoughts or questions?

goosemania: @quba42 , katello said they don't care about the deb_component for indexing, it means it's not critical what we do and even if there is valid pulp3 content or not, just that API needs to stay the same.

We might get away with a bit easier approach, let me talk to others.

quba42: I don't have any strong preferences (so long as it works).
I feel like I should be able to get this current approach to work, but if there is a better way of doing things.

The only other consideration I have not yet put a lot of thought into, is that in the pulp_deb case, the number of associated pulp3 content entries to a single pulp2 component unit can be very large for components like Debian main, or Ubuntu universe. My intuition is that lists of unit IDs, should still be OK efficiency wise, but I _could_ be wrong. Not sure if this is an argument against the serializer option.

goosemania: @quba42 , thinking a bit more about it and talking with @ipanova . We probably can't go with any simpler approach, even though Katello doesn't need the pulp2content mapping, we use it internally (I checked the code) to constract pulp3 repo versions, so we need pulp2content mapping to be in a good state anyway.

The serializer option and change to many2many is the last preferred option because it's risky and requires code change in many parts of the plugin.

That's the long way of saying that I suggest to continue with approach started in this PR.

quba42: I will go ahead and do that then. Thanks again for your time.

quba42:
As far as I am concerned, this can be closed if favor of #295. Of course, I have been wrong before. :wink:

goosemania:
Closed in favour of #295

PR #293, opened 2021-01-18
Fix null value in column "stream" error

dralley:


ipanova:
@dralley while we are at it, can we also fix the dependencies for Modules(they are optional as well)?

dralley:
Done

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -660,14 +660,20 @@ def _get_profiles(profiles):
...
+        for defaults in pulp2_content_batch:
+            kwargs = {
+                'module': defaults.name,
+                'digest': defaults.checksum,
+                'repo_id': defaults.repo_id,
+                'pulp2content': pulp2_id_obj_map[defaults.id]
+            }
+            if defaults.stream:
+                kwargs['stream'] = defaults.stream
+            if defaults.profiles:
```
ipanova, line 674: _get_profiles(defaults.profiles) if defaults.profiles else []

ipanova:
```
n [8]: Pulp2ModulemdDefaults(module='test2',pulp2content_id=p.pk).save()
---------------------------------------------------------------------------
NotNullViolation                          Traceback (most recent call last)
/usr/local/lib/pulp/lib64/python3.6/site-packages/django/db/backends/utils.py in _execute(self, sql, params, *ignored_wrapper_args)
     83             else:
---> 84                 return self.cursor.execute(sql, params)
     85 

NotNullViolation: null value in column "profiles" violates not-null constraint
DETAIL:  Failing row contains (80f2125a-b790-4de7-8ab1-98ef7362a636, 2021-01-14 17:23:34.868795+00, 2021-01-14 17:23:34.868838+00, test2, , null, , , 0329f8c7-e75b-4eb6-b39d-542ff6be7bc8).
```

dralley: I added it to the function instead.

ipanova: No what i meant is that in case profiles will be None  in defaults.profiles the kwargs will not have 'profiles' key at all so when saving pulp2modulemddefaults it will blow up with
`NotNullViolation: null value in column "profiles" violates not-null constraint`

dralley: What I mean is that I did this:

https://github.com/pulp/pulp-2to3-migration/pull/293/files#diff-669c91fe94bfb50613be9955c7cce760e8a0ee962b99c60544321bfd82783a4fR658

It's exactly equivalent to your suggestion just in a different place

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -655,19 +655,25 @@ def _get_profiles(profiles):
             """
 
             bson_string = BSON(profiles, encoding='utf8')
-            return bson_string.decode()
+            return bson_string.decode() or []
```
ipanova, line 658:
```
In [20]: profiles=None

In [21]: bson_string = BSON(profiles, encoding='utf8')
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-21-633a6001fe0d> in <module>
----> 1 bson_string = BSON(profiles, encoding='utf8')

TypeError: encoding without a string argument
```

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
ipanova: sorry i needed to be more explicit. I suggest removing ` if defaults.profiles:`
and have this in kwargs
```
            kwargs = {
                digest=defaults.checksum,
                'module': defaults.name,
                repo_id=defaults.repo_id,
                'digest': defaults.checksum,
                pulp2content=pulp2_id_obj_map[defaults.id])
                'repo_id': defaults.repo_id,
                'pulp2content': pulp2_id_obj_map[defaults.id]
               'profiles': _get_profiles(defaults.profiles) if defaults.profiles else []
}

```

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -612,7 +612,7 @@ class Pulp2ModulemdDefaults(Pulp2to3Content):
     # Unit key fields
     module = models.TextField()
     stream = models.TextField()
-    profiles = JSONField(dict)
```
dralley, line 615: I think maybe this should have been ```JSONField(default=dict)``` all along, and it got mixed up by accident.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
dralley: I think ```profiles``` is actually supposed to be a dict due to how it's stored in Pulp 2.  With the default value fixed properly, it seems like it works now.

If we're going to do a migration anyway I'm going to look for any other places where we might have made this mistake.

ipanova (approved):
changes seems good to me, ty for knocking this issue down.

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -761,6 +767,7 @@ def get_treeinfo_serialized(self):
             # Reset build_timestamp so Pulp will fetch all the addons during the next sync
             treeinfo_serialized['distribution_tree']['build_timestamp'] = 0
             return treeinfo_serialized
+        assert False, "Failed to find treeinfo, something went wrong"
```
ipanova, line 770: can we remove this line and open a separate pr for treeinfo related stuff? it does not seem to be related to the changes.

dralley: @ipanova Do you want the CI fix in a separate PR also?

PR #294, opened 2021-01-19
Don't generate sqlite metadata unless Pulp 2 config says so

dralley:
pulp_rpm provides a toggle for generating sqlite metadata, defaulting to
off, added in #7852. If the Pulp 2 config does specify generation of
sqlite metadata, then we need to re-enable it.

closes: #7851
https://pulp.plan.io/issues/7851

### DIFF
--- a/pulp_2to3_migration/app/plugin/rpm/repository.py
+++ b/pulp_2to3_migration/app/plugin/rpm/repository.py
@@ -73,7 +73,12 @@ class RpmDistributor(Pulp2to3Distributor):
             if pulp2_checksum_type:
                 checksum_types = {'metadata': pulp2_checksum_type,
                                   'package': pulp2_checksum_type}
-            publish(repo_version.pk, checksum_types=checksum_types)
+            sqlite = pulp2_config.get('generate_sqlite', False)
+            try:
+                publish(repo_version.pk, checksum_types=checksum_types, sqlite_metadata=sqlite)
+            except TypeError:
+                # hack, pulp_rpm <3.9 doesn't support sqlite_metadata kwarg
+                publish(repo_version.pk, checksum_types=checksum_types)
             publication = repo_version.publication_set.first()
 
         # create distribution
--- a/CHANGES/7851.bugfix
+++ b/CHANGES/7851.bugfix
@@ -1,0 +1,1 @@
+No longer generate sqlite metadata when publishing unless the Pulp 2 configuration specified to do so.


pulp_2to3_migration/app/plugin/rpm/repository.py
```diff
@@ -73,7 +73,8 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
             if pulp2_checksum_type:
                 checksum_types = {'metadata': pulp2_checksum_type,
                                   'package': pulp2_checksum_type}
-            publish(repo_version.pk, checksum_types=checksum_types)
+            sqlite = pulp2_config.get('generate_sqlite', False)
+            publish(repo_version.pk, checksum_types=checksum_types, sqlite_metadata=sqlite)
```
dralley, line 77: Problem: Only the master branch of ```pulp_rpm``` has this feature, which means that if we provide this flag, it's going to fail (unexpected keyword argument) with any currently-available release of the plugin.  I'm not sure of the best way to address this.

* We could backport this feature to older releases of pulp_rpm so that it will work properly
* We could backport a change to the the signature of ```publish(...)``` to include ```**kwargs``` so that it will be ignored

Neither of those are great options.  I don't think we can just break compatibility with any pulp_rpm except for an unreleased version since Katello won't be using it.

dralley: https://github.com/pulp/pulp_rpm/pull/1908/files#diff-e556922435d79157d5a5d7ebc182d611f4bcb3e441813d2266382ff6b15c9c8bL259

pulp_2to3_migration/app/plugin/rpm/repository.py
```diff
@@ -73,7 +73,8 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
             if pulp2_checksum_type:
                 checksum_types = {'metadata': pulp2_checksum_type,
                                   'package': pulp2_checksum_type}
-            publish(repo_version.pk, checksum_types=checksum_types)
+            sqlite = pulp2_config.get('generate_sqlite', False)
```
dralley, line 76: This should work right?  It's the same distributor config as https://github.com/pulp/pulp_rpm/blob/2-master/plugins/pulp_rpm/plugins/distributors/yum/publish.py#L1181 ?

pulp_2to3_migration/app/plugin/rpm/repository.py
goosemania: I think backporting a feature is not really an option since it's against all our rules :)
kwargs sounds like an option to me, the only pain is releasing 3.7 and 3.8.z releases.

I'm not sure which version of pulp_rpm Katello is using. They can upgrade to any until it's compatible with pulpcore 3.7.
Other options which are  not good or hacky but I throw them here anyway:
 - declare compatibility with pulpcore >= 3.9 that will force pulp_rpm version to be at least 3.9, they one with the feature in it. (it might create many inconveniences with packaging in downstream and with our development. Since  we still need to have compatibility with 3.7 pulpcore, just pulp_rpm needs to be 3.9. Are there conditional dependencies in pip world? :) )
 - hack with try/except to support both signatures

dralley: What Justin said:

> \<jsherrill> dalley: if pulp_rpm 3.9 is compatible with pulpcore 3.7, we can probably upgrade it in 3.7.  Assuming that 'upload' directory issue isn't a problem within pulp_rpm 3.7 (i assume thats a pulpcore issue?)
> \<jsherrill> err upgrade pulp_rpm to 3.9 i mean

goosemania: yeah, as expected, it doesn't help us though :)

ipanova: if understand the situation correctly, in case user has pulp_rpm 3.7, for example, no matter what we do, the sqlite will be generated, because pulp_rpm 3.7 code path does not any other option.
I would prefer a hack with the try/except to support both signatures and not fail. kwargs is less dirty approach however i don't know whether it is worth releasing 3.7.z and 3.8.z for the sake of this problem. I would image that in real world the user would anyway want to be on the latest release version before migrating all his data from pulp2 to pulp3.
And if the user is on older versions and does he not want the sqlite generated  he would not habe other choice then to upgrade first to pulp_rpm 3.9

ipanova:
This approach, works for me, lgtm.

ipanova (approved):

PR #295, opened 2021-01-21
Add the pulp2_subid field to the Pulp2Content model

quba42:
fixes #8137
https://pulp.plan.io/issues/8137

The intended use case is to provide one to many mapping from Pulp 2 to
Pulp 3 content. This is achieved by creating multiple Pulp2Content
entries that all share a single pulp2_id, and differ only via the
pulp2_subid.

This is needed for structured Debian migrations:
https://pulp.plan.io/issues/7865

The extra Pulp2Content subrecords are added by the relevant
pre_migrate_content_detail method. The reason is that it is only known
what subrecords are needed once the content is actually inspected by
this method.


quba42:
Note that the approach was taken from here: https://github.com/pulp/pulp-2to3-migration/pull/292

The new field is needed for: https://github.com/pulp/pulp-2to3-migration/pull/273

quba42:
Should I rename ` pulp_2to3_migration/app/migrations/0021_auto_20210121_1112.py` to something more meaningful?

quba42:
Not sure why this would cause test failures for RPM.

ipanova:
> Should I rename ` pulp_2to3_migration/app/migrations/0021_auto_20210121_1112.py` to something more meaningful?

yes please.

goosemania:
> Not sure why this would cause test failures for RPM.

Interesting, I wonder if it happens because every null is unique and is a part of uniqueness constraint.
Try to remove null=True and use an empty string as a default.

quba42:
> Interesting, I wonder if it happens because every null is unique and is a part of uniqueness constraint.
> Try to remove null=True and use an empty string as a default.

I found this in the Django docs:

> Avoid using null on string-based fields such as CharField and TextField because empty string values will always be stored as empty strings, not as NULL. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL.

goosemania:
@quba42 , your changes look good, thanks!
The CI is temperamental and we are looking at why it is so, it happens to many PRs, not only yours.

pulp_2to3_migration/app/models/content.py
```diff
@@ -33,9 +33,10 @@ class Pulp2Content(BaseModel):
     downloaded = models.BooleanField(default=False)
     pulp3_content = models.ForeignKey(Content, on_delete=models.SET_NULL, null=True)
     pulp2_repo = models.ForeignKey(Pulp2Repository, on_delete=models.SET_NULL, null=True)
+    pulp2_subid = models.CharField(max_length=255, blank=True)
```
ipanova, line 36: :tada:

quba42: Question: Should I send something on the mailing list that `blank=True` is preferable to `null=True` for string fields? I am pretty sure there are a lot of `null=True` string fields around the various Pulp repositories. How concerned should we be about this?

goosemania: I think it's always good to have such conversations on pulp-dev. I thought we had similar in the past but if we did it was looong ago. +1 to write to the mailing list and see why we didn't go that path in some cases.

quba42: Done: https://www.redhat.com/archives/pulp-dev/2021-January/msg00049.html

ipanova (approved):

goosemania (approved):

PR #297, opened 2021-01-24
Skip multi-artifact content properly when an artifact is unavailable.

dralley:
closes: #7469
https://pulp.plan.io/issues/7469


goosemania (approved):
Thanks! That will unblock further testing when artifacts are not available on the disc.

PR #298, opened 2021-01-26
Add batch_size to a pulp 2 query for content pre-migration

goosemania:
To avoid cursor timeout on a slow system.

[noissue]


ipanova:
@goosemania  is there any filed issue so we can reference it?

goosemania:
> @goosemania is there any filed issue so we can reference it?

No.

ipanova (approved):

dralley (approved):

PR #303, opened 2021-01-29
Taught rpm-migration to be more stable in the face of missing treeinfos.

ggainey:
fixes #8084

### DIFF
--- a/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
+++ b/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
@@ -741,7 +741,7 @@ class Pulp2Distribution(Pulp2to3Content):
         Create a Pulp 3 Distribution content for saving later in a bulk operation.
 
         Returns:
-            dict: a serialized treeinfo data
+            dict: a serialized treeinfo data, or None if no treeinfo can be found.
 
         """
         namespaces = [".treeinfo", "treeinfo"]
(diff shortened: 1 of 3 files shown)


pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -779,8 +778,11 @@ def create_pulp3_content(self):
 
         """
         treeinfo_serialized = self.get_treeinfo_serialized()
```
dralley, line 780: Should this have even been called if the content can't be created?  At minimum the docstring for ```get_treeinfo_serialized()``` needs to make it clear that it can fail.

ggainey: I'll add to the function-doc. DistributionTree-creation only happens if you have a treeinfo - you can't not-create it, until after you know whether get_treeinfo_serialized() worked (or not). This is just one more never-happen/edge-case where we flag a warning and continue, rather than abort the migration completely in the face of bad data (which is what happens w/out this fix, even if skip-corrupted=true)

goosemania:
@ggainey , so we won't fail with skip_corrupted=False, correct?
I'm not sure if it's the desired behaviour but we definitely can live with that.

goosemania (approved):

ggainey:
> @ggainey , so we won't fail with skip_corrupted=False, correct?
> I'm not sure if it's the desired behaviour but we definitely can live with that.

I believe it will follow the same path as any other 'corrupted' content follows. I will recheck today and respond here, just to be sure.

dralley (approved):

PR #304, opened 2021-02-02
Do not truncate taking system related tables between tests

goosemania:
It forces workers to restart and can casue various issues like task cancellation.

[noissue]


dralley (approved):
Yeah, that wouldn't be good :)

goosemania:
Tests still fail :) but less often :eyes:

PR #308, opened 2021-02-04
Update branch versions

goosemania:


ipanova (approved):

template_config.yml
```diff
@@ -2,14 +2,15 @@
...
 additional_plugins:
-- name: pulp_file
-  branch: master
-- name: pulp_container
-  branch: master
-- name: pulp_rpm
-  branch: master
-- name: pulp_deb
-  branch: main
+additional_plugins:
```
dralley, line 5: I think there's a typo here, ```additional_plugins``` on two lines.

goosemania: Interesting, good catch.
It's auto-generated though, I can look into plugin template maybe

goosemania: ah disregard, this is not.
It's my copy-paste.

PR #309, opened 2021-02-05
Fixed distributor re-run when it was only pre-migrated and then changed

goosemania:
The following situation is fixed now:
  * pre-migration for a distributor happened, but migration did not
  * then pulp2 distributor changed
  * then migration plan was run again

closes #8195
https://pulp.plan.io/issues/8195

### DIFF
--- a/CHANGES/8195.bugfix
+++ b/CHANGES/8195.bugfix
@@ -1,0 +1,1 @@
+Fixed distributor re-migration case when it was changed in Pulp 2 between migration plan runs.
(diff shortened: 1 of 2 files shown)


dralley (approved):

ipanova (approved):

ipanova:
I wonder how the first step happened? I believe the reporter did not mention any failures in the first migration run

goosemania:
@ipanova, any interruption in pre-migration step could cause that.
The steps I described was the only way to reproduce so far, so maybe user hasn't mentioned that or something else happened.

PR #311, opened 2021-02-08
Only go through the changed repositories on the re-run.

goosemania:
re #7779
https://pulp.plan.io/issues/7779

### DIFF
--- a/CHANGES/7779.bugfix
+++ b/CHANGES/7779.bugfix
@@ -1,0 +1,1 @@
+Fixed the re-run times when repositories/importers/distributors haven't changed much since the last run.
--- a/pulp_2to3_migration/app/models/base.py
+++ b/pulp_2to3_migration/app/models/base.py
@@ -33,6 +33,7 @@ def get_repo_types(plan):
     """
     repo_id_to_type = {}
     type_to_repo_ids = defaultdict(set)
+    is_simple_plan = False
 
     # mapping content type -> plugin/repo type, e.g. 'docker_blob' -> 'docker'
     content_type_to_plugin = {}
@@ -40,6 +41,12 @@ def get_repo_types(plan):
     for plugin in plan.get_plugin_plans():
         for content_type in plugin.migrator.pulp2_content_models:
             content_type_to_plugin[content_type] = plugin.migrator.pulp2_plugin
+
+        # if any of the plugin plans is empty, we'd need to go through all repo content relations
+        # to determine repo types correctly.
+        if plugin.empty:
+            is_simple_plan = True
+            continue
 
         repos = set(plugin.get_repositories())
         repos |= set(plugin.get_importers_repos())
@@ -49,22 +56,25 @@ def get_repo_types(plan):
             repo_id_to_type[repo] = plugin.type
         type_to_repo_ids[plugin.type].update(repos)
 
-    # TODO: optimizations.
-    # It looks at each content at the moment. Potential optimizations:
-    #  - This is a big query, paginate?
-    #  - Filter by repos from the plan
-    #  - Query any but one record for a repo
-    for rec in RepositoryContentUnit.objects().\
-            only('repo_id', 'unit_type_id').as_pymongo().no_cache():
-        repo_id = rec['repo_id']
-        unit_type_id = rec['unit_type_id']
+    # Go through repo content relations only when at least one of the plans is not complex,
+    # otherwise the type is determined by the plan in a much more efficient way.
+    if is_simple_plan:
+        # TODO: optimizations.
+        # It looks at each content at the moment. Potential optimizations:
+        #  - This is a big query, paginate?
+        #  - Filter by repos from the plan
+        #  - Query any but one record for a repo
+        for rec in RepositoryContentUnit.objects().\
+                only('repo_id', 'unit_type_id').as_pymongo().no_cache():
+            repo_id = rec['repo_id']
+            unit_type_id = rec['unit_type_id']
 
-        # a type for a repo is already known or this content/repo type is not supported
-        if repo_id in repo_id_to_type or unit_type_id not in content_type_to_plugin:
-            continue
-        plugin_name = content_type_to_plugin[unit_type_id]
-        repo_id_to_type[repo_id] = plugin_name
-        type_to_repo_ids[plugin_name].add(repo_id)
+            # a type for a repo is already known or this content/repo type is not supported
+            if repo_id in repo_id_to_type or unit_type_id not in content_type_to_plugin:
+                continue
+            plugin_name = content_type_to_plugin[unit_type_id]
+            repo_id_to_type[repo_id] = plugin_name
+            type_to_repo_ids[plugin_name].add(repo_id)
 
     return repo_id_to_type, type_to_repo_ids
 
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -1,6 +1,7 @@
 import logging
 
 from collections import namedtuple
+from datetime import datetime
 
 from django.db import transaction
 from django.db.models import Max, Q
@@ -279,11 +280,18 @@ def pre_migrate_all_without_content(plan):
     """
     Pre-migrate repositories, relations to their contents, importers and distributors.
 
-    NOTE: MongoDB and Django handle datetime fields differently. MongoDB doesn't care about
-    timezones and provides "naive" time, while Django is complaining about time without a timezone.
-    The problem is that naive time != time with specified timezone, that's why all the time for
-    MongoDB comparisons should be naive and all the time for Django/PostgreSQL should be timezone
-    aware.
+    Look at the last updated times in the pulp2to3 tables for repositories/importers/distributors:
+     * pulp2_last_unit_added or pulp2_last_unit_removed for repositories
+     * pulp2_last_updated for importers and distributors
+
+    Query empty-never-had-content repos (can't filter them out in any way) and repos for which
+    there were:
+     * content changes since the last run
+     * importer changes since the last run
+     * distributor changes since the last run
+
+    Query in order of last_unit_added for the case when pre-migration is interrupted before we are
+    done with repositories.
 
     Args:
         plan(MigrationPlan): A Migration Plan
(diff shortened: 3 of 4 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +355,46 @@ def pre_migrate_all_without_content(plan):
...
+            updated_impdist_repos = updated_imp_repos | updated_dist_repos
+            del updated_imp_repos
+            del updated_dist_repos
+
+            mongo_updated_repo_q = repo_id_q & (is_new_enough_repo_q | is_empty_repo_q)
+            mongo_updated_imp_dist_repo_q = mongo_Q(repo_id__in=updated_impdist_repos)
+
+            mongo_repo_qs = Repository.objects(
+                mongo_updated_repo_q | mongo_updated_imp_dist_repo_q
+            ).order_by('last_unit_added')
```
goosemania, line 397: no index in pulp2, maybe not a big deal since the number of repos is relatively low?

goosemania:
@dralley , @ggainey , @ipanova , this PR is far from being well tested, but I'd appreciate any thoughts or concerns on the approach.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +355,45 @@ def pre_migrate_all_without_content(plan):
             # filter by repo type
             repos_to_check = plan.type_to_repo_ids[plugin_plan.type]
 
-            mongo_repo_q = mongo_Q(repo_id__in=repos_to_check)
-            mongo_repo_qs = Repository.objects(mongo_repo_q)
+            # figure out which repositories/importers/distributors were updated since the last run
```
dralley, line 358: :frowning_face:

goosemania:
Approach
=======

Looking at the last updated times in the pulp2to3 tables for repositories/importers/distributors.
Query only empty repos (can't filter them out in any way) or repos for which:
 - there were any content changes
 - importer changes
 - distributor changes
Query in order of last_unit_added for the case when pre-migration is interrupted before we are done with repositories.

pulp_2to3_migration/app/pre_migration.py
ggainey: "How many is too many" may be difficult to set a hard limit on, since the failure mode is "query takes too much memory". We'll def want to experiment to see what it takes to fail this, I would think.

ggainey:
> # Approach

Just a note, I'd like to see this as a comment at the start of the code-section. I'm glad that my understanding after reading the code matches this - but it'd be WAY easier for the next person, to have this summary right next to the complicated scenario it describes.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +355,45 @@ def pre_migrate_all_without_content(plan):
...
+                repo_id_q & is_new_enough_imp_q
+            ).only('repo_id')
+            updated_imp_repos = set(imp.repo_id for imp in updated_importers)
+            updated_distributors = Distributor.objects(
+                repo_id_q & is_new_enough_dist_q
+            ).only('repo_id')
+            updated_dist_repos = set(dist.repo_id for dist in updated_distributors)
+            updated_impdist_repos = updated_imp_repos | updated_dist_repos
+            del updated_imp_repos
+            del updated_dist_repos
```
dralley, line 389: ?

pulp_2to3_migration/app/pre_migration.py
dralley: Probably fine.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +355,45 @@ def pre_migrate_all_without_content(plan):
             # filter by repo type
             repos_to_check = plan.type_to_repo_ids[plugin_plan.type]
 
-            mongo_repo_q = mongo_Q(repo_id__in=repos_to_check)
-            mongo_repo_qs = Repository.objects(mongo_repo_q)
+            # figure out which repositories/importers/distributors were updated since the last run
+            repo_premigrated_last = Pulp2Repository.objects.aggregate(
+                Max('pulp_last_updated')
```
ipanova, line 360: i don't think we have last_updated field on the repository model

ipanova: nevermind me

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +355,45 @@ def pre_migrate_all_without_content(plan):
...
+            )['pulp_last_updated__max'] or 0
+            dist_premigrated_last = Pulp2Distributor.objects.aggregate(
+                Max('pulp_last_updated')
+            )['pulp_last_updated__max'] or 0
+
+            is_content_added_q = mongo_Q(last_unit_added__gte=repo_premigrated_last)
+            is_content_removed_q = mongo_Q(last_unit_removed__gte=repo_premigrated_last)
+            is_new_enough_repo_q = is_content_added_q | is_content_removed_q
+            is_empty_repo_q = (
+                mongo_Q(last_unit_added__exists=False) & mongo_Q(last_unit_removed__exists=False)
```
ipanova, line 373: well, the best way would be to check for content_unit_count which we do not use as of now in the migration plugin.
you might still have a repo with both of these fields set but i believe they would get captures on lines 369 and 370?

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +355,45 @@ def pre_migrate_all_without_content(plan):
...
+            )
+            is_new_enough_imp_q = mongo_Q(last_updated__gte=imp_premigrated_last)
+            is_new_enough_dist_q = mongo_Q(last_updated__gte=dist_premigrated_last)
+            repo_id_q = mongo_Q(repo_id__in=repos_to_check)
+
+            updated_importers = Importer.objects(
+                repo_id_q & is_new_enough_imp_q
+            ).only('repo_id')
+            updated_imp_repos = set(imp.repo_id for imp in updated_importers)
```
ipanova, line 382: what is the need of making a set out of it?

pulp_2to3_migration/app/pre_migration.py
ipanova: can you please explain why do we need to trough empty repos?

goosemania: > well, the best way would be to check for content_unit_count which we do not use as of now in the migration plugin.
> you might still have a repo with both of these fields set but i believe they would get captures on lines 369 and 370?

Yes, the empty ones which were not empty at some point can be filtered out, so I think handling them on 369/370 is a benefit.

goosemania: > can you please explain why do we need to trough empty repos?

Just posting my IRC reply here for those who missed it.

Empty repos do not have any timestamp associated with them, and we need to make sure that they are part of the query because we need to migrate them. There is no way to know whether we migrated them or not by looking at pulp2 db only, so they are always a part of a pulp2 queryset.

pulp_2to3_migration/app/pre_migration.py
goosemania: Probably no need. We need to have a sequence containing repo_id and a set is a safe way to ensure they are unique.
I agree that importers are not reused in pulp2, so should be unique regardless. I can switch it to tuple or list if it's preferable.

pulp_2to3_migration/app/pre_migration.py
goosemania: Just to get rid of extra copies of data  which is not needed anymore. I can keep it.

dralley: It's not a problem.  If it's for memory purposes, it won't actually help.  If it's just to make it impossible to use the old copy by accident then it's worthwhile.

pulp_2to3_migration/app/pre_migration.py
ipanova: we can just .distinct('repo_id') this will give you a list with unique rpeo_ids

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -355,8 +362,47 @@ def pre_migrate_all_without_content(plan):
...
+            )['pulp2_last_updated__max'] or epoch
+            dist_premigrated_last = Pulp2Distributor.objects.aggregate(
+                Max('pulp2_last_updated')
+            )['pulp2_last_updated__max'] or epoch
+
+            is_content_added_q = mongo_Q(last_unit_added__gte=repo_premigrated_last_by_added)
+            is_content_removed_q = mongo_Q(last_unit_removed__gte=repo_premigrated_last_by_removed)
+            is_new_enough_repo_q = is_content_added_q | is_content_removed_q
+            is_empty_repo_q = (
+                mongo_Q(last_unit_added__exists=False) & mongo_Q(last_unit_removed__exists=False)
```
goosemania, line 384: It might be enough to check last_unit_added__exists=False only. I guess no units can be removed if they hadn't been added to a repo in the first place.

CHANGES/7779.bugfix
```diff
@@ -0,0 +1 @@
+Fixed the re-run times when repositories/importers/distributors hasn't changed much since the last run.
```
ipanova, line 1: s/hasn't/haven't/

pulp_2to3_migration/app/pre_migration.py
ipanova: let's do that. other repos that used to contain content and then it got removed would be handled on the upper lines.

pulp_2to3_migration/app/pre_migration.py
ipanova: i am still suggesting to use distinct in this case and not iterating through the qs
```
d=Distributor.objects.distinct('repo_id')
i=Importer.objects.distinct('repo_id')
result = set().union(d,i)
```

ipanova:
The PR looks good to me. I would add comments on some querysets, while reading the code i need to go back up and re-fresh my memory about what `is_new_enough_repo_q ` contains, for example.

pulp_2to3_migration/app/pre_migration.py
goosemania: @ipanova , I'm afraid of mongo's distinct, it can cause OOM easily, not sure how much data is needed for it. I experienced it in the past, with applicability specifically. I'd do as little heavy operations as possible in mongo anyway since it's running in production and loaded with other stuff.

ipanova: @goosemania that's a fair point.

goosemania:
This improved a huge setup migration only by 25% which is not good enough, so I continue working on it

pulp_2to3_migration/app/pre_migration.py
ggainey: This code ran against a test-server with many (tens of thousands) of repos and didn't fail. I think we're good without worrying about a Pulp2 index.

ggainey:
> This improved a huge setup migration only by 25% which is not good enough, so I continue working on it

The most-recent pushed changes took the migration-setup from 8.5 hours down to 1. There may be more we can do, but I think an order of magnitude is a great place to start.

ggainey (approved):
This code had a 10x impact on our worst-case server migration. I approve; I'd like to see it accepted, and if there's more to do in this area, let's do it in a new issue/PR.

pulp_2to3_migration/app/models/base.py
```diff
@@ -49,22 +56,25 @@ def get_repo_types(plan):
...
-
-        # a type for a repo is already known or this content/repo type is not supported
-        if repo_id in repo_id_to_type or unit_type_id not in content_type_to_plugin:
-            continue
-        plugin_name = content_type_to_plugin[unit_type_id]
-        repo_id_to_type[repo_id] = plugin_name
-        type_to_repo_ids[plugin_name].add(repo_id)
+    # Go through repo content relations only when at least one of the plans is not complex,
+    # otherwise the type is determined by the plan in a much more efficient way.
+    if is_simple_plan:
```
dralley, line 61: Can we exclude all of the repositories we do know the types of?

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -296,20 +304,62 @@ def pre_migrate_all_without_content(plan):
...
+            updated_impdist_repos = updated_imp_repos | updated_dist_repos
+
+            mongo_updated_repo_q = repo_repo_id_q & (is_new_enough_repo_q | is_empty_repo_q)
+            mongo_updated_imp_dist_repo_q = mongo_Q(repo_id__in=updated_impdist_repos)
+
+            mongo_repo_qs = Repository.objects(
+                mongo_updated_repo_q | mongo_updated_imp_dist_repo_q
+            ).order_by('last_unit_added')
+
+            pb.total += mongo_repo_qs.count()
```
dralley, line 361: In the code below I would recommend putting the ```pb.increment()``` inside the transaction.

Also it doesn't look like ```notes``` is used by the function below, or by any of the ```pre_migrate_*``` functions, maybe we can drop it from the query.

dralley (approved):
This looks good, I just left some suggestions about unrelated code.

PR #312, opened 2021-02-11
Removed no-longer-needed pre_migration codepath.

ggainey:
Removes part of the changes for #7280.

fixes #8166

### DIFF
--- a/CHANGES/8166.bugfix
+++ b/CHANGES/8166.bugfix
@@ -1,0 +1,1 @@
+Fixed an edge-case failure in erratum-migration when doing repeated migrations.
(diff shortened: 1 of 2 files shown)


goosemania:
:tada:

goosemania:
Could we merge it? I believe we are past the pre-migration point in the dogfood testing

ggainey:
> Could we merge it? I believe we are past the pre-migration point in the dogfood testing

Good idea - we've tested that it fixes the fatal failure successfuly. We'll need to pay attention to subtle finished-migration issues, but anything found should get a new issue.

goosemania (approved):

PR #321, opened 2021-02-22
Remove unused code paths for a simple plan

goosemania:
closes #8288
https://pulp.plan.io/issues/8288

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -369,15 +369,12 @@ def pre_migrate_all_without_content(plan):
                 repo = None
                 repo_id = repo_data.repo_id
                 with transaction.atomic():
-                    if not repos or repos and repo_id in repos:
+                    if repo_id in repos:
                         repo = pre_migrate_repo(repo_data, plan.repo_id_to_type)
-                    # do not pre-migrate importers/distributors in case of special repo setup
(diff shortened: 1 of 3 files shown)


goosemania:
I do not think I removed anything useful but I'd appreciate another pair of eyes to go through it thoroughly.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -625,12 +604,9 @@ def handle_outdated_resources(plan):
     for plugin_plan in plan.get_plugin_plans():
         inplan_repos = plugin_plan.get_repositories()
```
dralley, line 605: Should we make this a set to begin with?  I think it's a list, but I'm not sure of any particular reason why it needs to be a list, and it might have duplicate entries for some IDs.

dralley (approved):

pulp_2to3_migration/app/pre_migration.py
goosemania: Yeah, we can make it a tuple, technically it would be an invalid/wrong plan if it had duplicated repo ids listed.
This list is specifically for repositories and it doesn't include importer or distributor ones.

dralley: Not targeted specifically at the repository list - actually it's the importer and distributor lists where it's more obvious how duplicates might sneak in.

PR #322, opened 2021-02-22
Fix re-migration when distributors/importers changed in migration plan

goosemania:


pulp_2to3_migration/app/tasks/reset.py
```diff
@@ -33,6 +34,7 @@ def reset_pulp3_data(migration_plan_pk):
                    total=len(pulp2_plugins))
     with ProgressReport(**pb_data) as pb:
         for plugin in plan.get_plugin_plans():
+            RepoSetup.reset_plugin(plugin.type)
```
goosemania, line 37: I haven't added any tests for the reset. I'm not sure what to test apart from what we've already had for the reset feature.

dralley:
@goosemania Do you know if this strategy could be used to fix https://pulp.plan.io/issues/5981 ?

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -401,3 +401,95 @@ def test_no_immediate_importer(self):
 
         self.assertEqual(self.file_repo_versions_api.list(pulp3_repo.pulp_href).count, 2)
         self.assertEqual(repo_content.count, 3)
+
+    def test_importer_swap(self):
+        """
+        Test that if only importers in migration plan changed, the changes are noticed.
+
+        Importers themselves haven't changed in Pulp 2, only where they are specified in the
```
ipanova, line 409: should we add a test when there has been a swap and yet one of the importers/distributors have changed in pulp2?

I know that we have a separate test that ensures importers are re-migrated if feed changes and similar for distributors config i just wonder if it makes sense to combine both cases together. I will come back to this suggestion tomorrow morning ( maybe it's overkill)

goosemania: I think I can add such test if we have already a snapshot with the changes. I think we do.

goosemania:
> @goosemania Do you know if this strategy could be used to fix https://pulp.plan.io/issues/5981 ? It still reproduces

I would think so. Name will be another resource type.

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -401,3 +401,95 @@ def test_no_immediate_importer(self):
...
+        Migration Plan.
+        """
+        # run for the first time with the standard plan
+        self._load_and_run(FILE_COMPLEX_PLAN)
```
dralley, line 413: Would importer swap impact LCEs at all, and if so, does the ISO portion of the RPM plugin utilize them in the same way as RPM?

goosemania: Swap wouldn't impact them because LCEs are handled in relation to content. The only thing that matters is that the importer is being migrated but doesn't matter which repo it relates to.
I believe that ISO LCEs and RPM LCEs are used and handled in the same way.
Great questions!

dralley (approved):

pulp_2to3_migration/app/models/base.py
```diff
@@ -330,7 +337,153 @@ def _parse_plugin_plan(self, repository_data):
...
+                # ignore existing relations, they've been already updated as up to date ones
+                pass
+
+    @classmethod
+    def mark_changed_relations(cls):
+        """
+        Set is_migrated to False for any relations which changed and no longer up to date.
+        """
+        changed_relations_repo_ids = RepoSetup.objects.exclude(
+            status=cls.UP_TO_DATE
```
goosemania, line 485: @ipanova , this is a fix for the "moved" case, it used to mark only `new` ones and didn't look at `old` relations.

ipanova (approved):
looks good :clinking_glasses: :cheese:

PR #326, opened 2021-03-03
Adds notes about worker timeouts in config docs

bmbouter:
closes #7847

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -8,6 +8,11 @@ Requirements
 * access to Pulp 2 database
 * [recommended] all Pulp 3 supported checksum types are allowed
 
+.. note::
+
+    The migration workload is very I/O intensive. It's highly recommended that the volume
+    backing your Postgresql database be on a SSD or other low latency, high IOPS storage volume.
+
 
 Configuration
 -------------
@@ -19,6 +24,13 @@ On Pulp 2 machine
 one of the ``bindIP`` in ``/etc/mongod.conf``.
 
 2. Make sure ``/var/lib/pulp`` is on a shared filesystem.
+
+.. note::
+
+    If you experience Pulp 2 workers timing out during the migration, consider making them more
+    tolerant to delayed mongodb writes by increasing the ``worker_timeout`` setting in the
+    ``[tasks]`` section of your Pulp 2 ``server.conf``. Typically a value of 300 seconds has worked
+    well even during large migration workloads.
 
 
 On Pulp 3 machine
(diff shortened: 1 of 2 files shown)


docs/configuration.rst
```diff
@@ -8,6 +8,11 @@ Requirements
 * access to Pulp 2 database
 * [recommended] all Pulp 3 supported checksum types are allowed
 
+.. note::
+
+    The migration workload is very SQL-write intensive. It's highly recommended that the volume
```
dralley, line 13: I would just call it IO-intensive.  We know about the worker timeout problem due to SQL writes, but it does a lot of reading and file ops as well, which could influence the system's performance.

docs/configuration.rst
```diff
@@ -63,4 +75,10 @@ from the setting.
...
 the setting, you likely will need to run ``pulpcore-manager handle-artifact-checksums`` to remove
-unsupported checksums from the database, or Pulp will refuse to start.
\ No newline at end of file
+unsupported checksums from the database, or Pulp will refuse to start.
+
+.. note::
+
+    If you experience Pulp 3 workers timing out during the migration, consider making them more
+    tolerant to delayed PostgreSQL writes by increasing the Pulp 3 ``WORKER_TTL`` setting. Typically
+    a value of 300 seconds has worked well even during large migration workloads.
```
dralley, line 84: Depending on whether we backport this patch, a note about which Pulpcore version is required to do this might make sense.

goosemania: +1 to mention 3.11. 
I believe we do not plan to backport it. Katello will likely patch the hardcoded constant we used to have in any older releases they need.

bmbouter: I'll add the 3.11 reference. We can't backport in pulpcore because it's a feature.

bmbouter:
@dralley and @goosemania this is ready for review again I think. Lmk what you think :)

docs/configuration.rst
```diff
@@ -63,4 +75,11 @@ from the setting.
...
-unsupported checksums from the database, or Pulp will refuse to start.
\ No newline at end of file
+unsupported checksums from the database, or Pulp will refuse to start.
+
+.. note::
+
+    If you experience Pulp 3 workers timing out during the migration, consider making them more
+    tolerant to delayed PostgreSQL writes by increasing the Pulp 3 ``WORKER_TTL`` setting. Typically
+    a value of 300 seconds has worked well even during large migration workloads. This setting is
+    available in ``pulpcore>=3.11``.
```
goosemania, line 85: :+1:

docs/configuration.rst
```diff
@@ -8,6 +8,11 @@ Requirements
 * access to Pulp 2 database
 * [recommended] all Pulp 3 supported checksum types are allowed
 
+.. note::
+
+    The migration workload is very I/O intensive. It's highly recommended that the volume
+    backing your Postgresql database be on a SSD or other low latency, high IOPs storage volume.
```
goosemania, line 14: Should it be `IOPS`? I'm not sure but I think I recall seeing it in uppercase only.

goosemania (approved):
I left one minor cooment.
Thanks for the docs and most importantly for all the investigation in this area.

docs/configuration.rst
bmbouter: It should be `IOPS` since second is part of the acronym. :+1:

dralley (approved):

PR #328, opened 2021-03-04
Refactor file tests so it is easier to add more.

goosemania:
[noissue]

### DIFF
--- a/pulp_2to3_migration/tests/functional/file_base.py
+++ b/pulp_2to3_migration/tests/functional/file_base.py
@@ -1,0 +1,83 @@
+import time
+
+from pulpcore.client.pulpcore import Configuration
+from pulpcore.client.pulp_file import (
+    ApiClient as FileApiClient,
+    ContentFilesApi,
+    DistributionsFileApi,
+    PublicationsFileApi,
+    RemotesFileApi,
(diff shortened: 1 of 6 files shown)


goosemania:
I'm adding tests for #322 and it was very inconvenient to do. So I did the same what we have for rpm, created a base class for tests to reduce copy paste. In a separate PR, to extract irrelevant changes from #322 .

pulp_2to3_migration/tests/functional/file_base.py
```diff
@@ -0,0 +1,85 @@
...
+from pulp_smash import cli
+from pulp_smash import config as smash_config
+from pulp_smash.pulp3.bindings import monitor_task, monitor_task_group
+
+from .constants import BINDINGS_CONFIGURATION, TRUNCATE_TABLES_QUERY_BASH
+
+
+class BaseTestFile:
+    """
+    Test RPM migration re-runs.
```
ipanova, line 31: s/RPM/file or iso?

ipanova (approved):

pulp_2to3_migration/tests/functional/file_base.py
goosemania: copy-pasta

PR #331, opened 2021-03-09
Fixes a Pulp2Content unique-constraint that wasn't very unique.

ggainey:
In Postgres, all NULLs are unique - so if a nullable field is included
in a UiniqueConstraint, it allows rows that differ only by the NULL.

Build two constraints, one when the 'optional' field is NULL, and
one where it has content.

fixes #8329
[nocoverage]


pulp_2to3_migration/app/migrations/0024_errata_to_subid.py
```diff
@@ -0,0 +1,33 @@
...
+    for dup in p2content:
+        # ...find the pulp-ids of entries in that set, order by pulp2-last-update
+        to_delete = pulp2content.objects.filter(**{x: dup[x] for x in ['pulp2_id', 'pulp2_content_type_id', 'pulp2_repo_id', 'pulp2_subid']}).order_by('pulp2_last_updated')
+        # exclude the first/newest one...
+        to_delete = to_delete.exclude(pk=to_delete[0].pk)
+        # ...and delete the rest
+        to_delete.delete()
+
+    # for existing errata, copy repo_id to subid to maintain uniqueness
+    errata = pulp2content.objects.filter(pulp2_content_type_id='erratum').update(pulp2_subid=F('pulp2_repo_id'))
```
ggainey, line 23: With the 'correct' UQ (if repo-id is null we demand that p2-id/p2-content-type-id/p2-subid is unique), we may not need to duplicate repo-id into subid. Leaving this here awaiting further discussion.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -178,7 +178,8 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
                     pulp2_last_updated=record._last_updated,
                     pulp2_storage_path=record._storage_path,
                     downloaded=downloaded,
-                    pulp2_repo=relation.pulp2_repository
+                    pulp2_repo=relation.pulp2_repository,
+                    pulp2_subid=relation.pulp2_repository.pk,  # May not need this with correct UQ
```
ggainey, line 182: See discussion under migration 0024. 

Assuming we decide we still need to fill subid with repo_id for errata: this code is not plugin-aware and can't have an "if this is an erratum" switch.  The only reason what I'm doing here is OK is because the only thing that sets set_pulp2_repo to true is Pulp2Erratum. I don't like implicit assumptions. We can add a set_subid flag, we can rename set_repo_id to set_repo_and_sub_id, we can rename it to set_sub_id and drop the repo_id column.

ggainey:
This is in draft for two reasons:

First - Making the UQ 'correct' means we probably do NOT need to make sure errata store repo_id in subid. After discussion around this approach, we may remove the Pulp2Content(type=erratum).update() from the migration, and the code that copies repo-id into subid.

Second, and more painful - with the corrected UQ in place, the reproducer from https://pulp.plan.io/issues/8329#note-3 fails with the stacktrace noted in comment-10. Discussion needed on how we address the problem.

pulp_2to3_migration/app/migrations/0025_errata_to_subid.py
```diff
@@ -0,0 +1,33 @@
...
+def fix_dups(apps, schema_editor):
+    pulp2content = apps.get_model("pulp_2to3_migration", "Pulp2Content")
+    # find sets of p2-id/p2-type/-2-repo/p2-sub dups
+    p2content = pulp2content.objects.values('pulp2_id', 'pulp2_content_type_id', 'pulp2_repo_id', 'pulp2_subid').order_by('pulp2_id')
+    p2content = p2content.annotate(count_id=models.Count("pulp2_id")).filter(count_id__gt=1)
+
+    # for each set of dups...
+    for dup in p2content:
+        # ...find the pulp-ids of entries in that set, order by pulp2-last-update
+        to_delete = pulp2content.objects.filter(**{x: dup[x] for x in ['pulp2_id', 'pulp2_content_type_id', 'pulp2_repo_id', 'pulp2_subid']}).order_by('pulp2_last_updated')
```
ggainey, line 16: Need to reverse the sense of the order-by to get the newest/largest-date as the first entry

ggainey: And move dict-expansion to its own line to be more clear what it's doing

pulp_2to3_migration/app/migrations/0024_errata_to_subid.py
ggainey: After discussion agreement that we don't need to touch errata-entries - removed.

ggainey:
> This is in draft for two reasons:
> 
> First - Making the UQ 'correct' means we probably do NOT need to make sure errata store repo_id in subid. After discussion around this approach, we may remove the Pulp2Content(type=erratum).update() from the migration, and the code that copies repo-id into subid.
> 
> Second, and more painful - with the corrected UQ in place, the reproducer from https://pulp.plan.io/issues/8329#note-3 fails with the stacktrace noted in comment-10. Discussion needed on how we address the problem.

Both of these issues should now be resolved.

pulp_2to3_migration/app/models/content.py
```diff
@@ -36,7 +38,15 @@ class Pulp2Content(BaseModel):
...
     class Meta:
-        unique_together = ('pulp2_id', 'pulp2_content_type_id', 'pulp2_repo', 'pulp2_subid')
+        constraints = [
+            UniqueConstraint(
+                fields=['pulp2_id', 'pulp2_content_type_id', 'pulp2_repo', 'pulp2_subid'],
+                name='unique_with_optional'
+            ),
+            UniqueConstraint(fields=['pulp2_id', 'pulp2_content_type_id', 'pulp2_subid'],
+                             condition=Q(pulp2_repo=None),
+                             name='unique_without_optional'),
```
goosemania, line 48: :+1:

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -211,6 +211,25 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
...
                                                                   ignore_conflicts=True)
+
+            # bulk_create(ignore_conflicts=True) hands back the same item-set we passed in,
+            # *even if* it decided to update an existing db-record rather than creating a new
+            # one with the passed-in PK. As a result, we can't trust pulp2content_batch to
+            # have the 'right' PKs (i.e., the in-memory p2content_batch doesn't match the
+            # db-reality). This causes the pre_migrate_content_detail() below to fail as it
+            # attempts to create detail-records for the Pulp2Content records it's been handed.
+            # THEREFORE - we need to find the 'real' IDs of everything in p2content-batch based
+            # on its uniqueness-fields and update the in-memory list with them.
```
goosemania, line 222: Thanks for detailed comment.

goosemania (approved):
Thank you! :octocat:

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -211,6 +211,25 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
...
+            for p2c in pulp2content_batch:
+                filter_q = Q(
+                    pulp2_content_type_id=content_type,
+                    pulp2_id=p2c.pulp2_id,
+                    pulp2_repo=p2c.pulp2_repo,
+                    pulp2_subid=p2c.pulp2_subid,
```
dralley, line 228: I don't know if this is actually the case, but filtering on a *maybe* unindexed field (unindexed for everything other than errata) might cause perf degredation.

pulp_2to3_migration/app/migrations/0025_remove_p2content_dups.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def fix_dups(apps, schema_editor):
+    pulp2content = apps.get_model("pulp_2to3_migration", "Pulp2Content")
+    # find sets of p2-id/p2-type/-2-repo/p2-sub dups
+    p2content = pulp2content.objects.values('pulp2_id', 'pulp2_content_type_id', 'pulp2_repo_id', 'pulp2_subid').order_by('-pulp2_id')
+    p2content = p2content.annotate(count_id=models.Count("pulp2_id")).filter(count_id__gt=1)
+
+    # for each set of dups...
+    for dup in p2content:
```
dralley, line 13: ```.iterator()```

dralley: I guess it's not a huge deal unless there's a TON of duplicates

goosemania: it should be very few duplicates if any. It happens when pulp2 content (not errata) changes and it's only for some fields like `downloaded`.

ggainey: Also, the only people who can be in this situation right now is "us". I'm OK with leaving it like this.

pulp_2to3_migration/app/pre_migration.py
ggainey: pulp_deb uses it as well. Also, this code can't "know" anything about plugins - it only knows that these 4 fields are the 'natural keys' of Pulp2Content, and we can't rely on PK. So, here we are :( 

This whole loop is a performance hit. Unfortunately, the behavior of bulk_create() leaves us no other choice. Slow is bad - but FK-violation-exceptions are way worse.

PR #334, opened 2021-03-11
Fix migration for the case when no importer is specified.

goosemania:
closes #8382
https://pulp.plan.io/issues/8382

### DIFF
--- a/pulp_2to3_migration/app/models/base.py
+++ b/pulp_2to3_migration/app/models/base.py
@@ -434,7 +434,7 @@ class RepoSetup(BaseModel):
             pulp2_resource_type=cls.IMPORTER,
             pulp2_repo_type=repo_type,
             pulp2_repo_id=repo_id,
-            pulp2_resource_repo_id=importer_repo_id,
+            pulp2_resource_repo_id=importer_repo_id or '',
             defaults={'status': cls.NEW}
         )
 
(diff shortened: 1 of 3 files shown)


pulp_2to3_migration/app/models/base.py
```diff
@@ -478,7 +478,7 @@ def set_distributors(cls, repo_id, repo_type, distributor_repo_ids):
                     pulp2_resource_type=cls.DISTRIBUTOR,
                     pulp2_repo_type=repo_type,
                     pulp2_repo_id=repo_id,
-                    pulp2_resource_repo_id=distributor_repo_id,
+                    pulp2_resource_repo_id=distributor_repo_id or '',
```
goosemania, line 481: I do not believe it's realistic to have no distributors but I added just in case.

pulp_2to3_migration/tests/functional/test_migration_behaviour.py
```diff
@@ -303,3 +320,19 @@ def test_no_immediate_importer(self):
 
         self.assertEqual(self.file_repo_versions_api.list(pulp3_repo.pulp_href).count, 2)
         self.assertEqual(repo_content.count, 3)
+
+    def test_no_importer(self):
```
ggainey, line 324: yay for tests!

pulp_2to3_migration/app/models/base.py
ggainey: Is there anywhere else we set this that might have the same issue? or just the two uses in this file?

ggainey: N<, I did my own search, we're good :)

ggainey (approved):
looks good to me

goosemania:
> Is there anywhere else we set this that might have the same issue? or just the two uses in this file?

Only here, other places use those methods.

PR #337, opened 2021-03-12
Switch back to ubuntu 18

fao89:
GHA has switched from ubuntu 18 to 20,
the change is beyond the distro, some packages are not installed for example.
- https://github.com/actions/virtual-environments/blob/ubuntu18/20210302.0/images/linux/Ubuntu1804-README.md
- https://github.com/actions/virtual-environments/blob/ubuntu20/20210302.0/images/linux/Ubuntu2004-README.md

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ on:
 jobs:
 
   lint:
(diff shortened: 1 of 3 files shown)


fao89:
the issue is not related

PR #341, opened 2021-03-22
Remove Pulp2Content records if content is no longer in Pulp2

goosemania:
If some content is only pre-migrated and not migrated to Pulp 3,
and then removed from Pulp 2, it can cause a problem with subsequent migration run.
Such content is treated as corrupted but in reality, it's no longer in Pulp 2.
Now such records are removed before any pre-migration starts.

closes #7887
https://pulp.plan.io/issues/7887

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -77,11 +77,39 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
         content_model: Models for content which is being migrated.
         mutable_type: Boolean that indicates whether the content type is mutable.
     """
+    def delete_removed_pulp2_content(content_model):
+        """
+        Delete Pulp2Content records for content which is no longer present in Pulp2.
+
+        This is to avoid situations and extra work when not all content migrated during the first
+        migration run, then orphan clean up is run in Pulp 2, and then migration is run again.
+
+        Args:
+            content_model: Pulp 2 content model
+
+        """
+        content_type = content_model.pulp2.TYPE_ID
+        mongo_content_qs = content_model.pulp2.objects().only('id')
+        mongo_content_ids = {c['_id'] for c in mongo_content_qs.as_pymongo().no_cache()}
+        premigrated_content_ids = set(
+            Pulp2Content.objects.filter(
+                pulp2_content_type_id=content_type
+            ).only('pulp2_id').values_list('pulp2_id', flat=True)
+        )
+        content_ids_to_delete = premigrated_content_ids - mongo_content_ids
+        if content_ids_to_delete:
+            Pulp2Content.objects.filter(
+                pulp2_content_type_id=content_type,
+                pulp2_id__in=content_ids_to_delete
+            ).delete()
+
     batch_size = 100
     pulp2content = []
     pulp2mutatedcontent = []
     content_type = content_model.pulp2.TYPE_ID
     set_pulp2_repo = content_model.pulp_2to3_detail.set_pulp2_repo
+
+    delete_removed_pulp2_content(content_model)
 
     # the latest timestamp we have in the migration tool Pulp2Content table for this content type
     content_qs = Pulp2Content.objects.filter(pulp2_content_type_id=content_type)
--- a/pulp_2to3_migration/tests/functional/test_rpm_rerun.py
+++ b/pulp_2to3_migration/tests/functional/test_rpm_rerun.py
@@ -189,6 +189,7 @@ class BaseTestRpmRerun(BaseTestRpm):
                 'remotes': defaultdict(dict),
                 'publications': defaultdict(dict),
                 'distributions': defaultdict(dict),
+                'pulp2content': defaultdict(dict),
             }
             for repo in cls.rpm_repo_api.list().results:
                 latest_version = cls.rpm_repo_versions_api.read(repo.latest_version_href)
@@ -209,6 +210,11 @@ class BaseTestRpmRerun(BaseTestRpm):
                 data['distributions'][dist.name] = {
                     'created': dist.pulp_created
                 }
+            for pulp2content in cls.pulp2content_api.list().results:
+                data['pulp2content'][pulp2content.pulp2_id] = {
+                    'pulp3content': pulp2content.pulp3_content
+                }
+
             return data
 
         set_pulp2_snapshot(name='rpm_base_4repos')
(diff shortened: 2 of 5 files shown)


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -77,12 +77,39 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
...
+        premigrated_content_ids = set(
+            Pulp2Content.objects.filter(
+                pulp2_content_type_id=content_type
+            ).only('pulp2_id').values_list('pulp2_id', flat=True)
+        )
+        content_ids_to_delete = premigrated_content_ids - mongo_content_ids
+        Pulp2Content.objects.filter(
+            pulp2_content_type_id=content_type,
+            pulp2_id__in=content_ids_to_delete
+        ).delete()
```
goosemania, line 103: Instead, I could use `exclude`:
`Pulp2Content.objects.filter(pulp2_content_type_id=content_type).exclude(pulp2_id__in=mongo_content_ids).delete()`

But I'm not sure if it's better to pass a huge list of ids into exclude. Is there a limit on the query size?
Likely, in the majority of cases, `mongo_content_ids` list is big and the `content_ids_to_delete` is small.

dralley (approved):

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -77,12 +77,39 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
...
+            content_model: Pulp 2 content model
+
+        """
+        content_type = content_model.pulp2.TYPE_ID
+        mongo_content_qs = content_model.pulp2.objects().only('id')
+        mongo_content_ids = {c['_id'] for c in mongo_content_qs.as_pymongo().no_cache()}
+        premigrated_content_ids = set(
+            Pulp2Content.objects.filter(
+                pulp2_content_type_id=content_type
+            ).only('pulp2_id').values_list('pulp2_id', flat=True)
```
dralley, line 97: I think the .only is unnecessary if you do .values_list

pulp_2to3_migration/app/pre_migration.py
dralley: Using subqueries is almost always better than passing lists of IDs, if it can be helped.  Granted, this is Mongo, I'm less familiar with how it would behave.

goosemania: Mongo is only on line 92-93, and we get list of ids from mongo anyway, so we can't avoid passing it.
But for Pulp2Content (which is postgres), is it better to pass once a huge list, or make couple of queries and pass a smaller one later?

dralley: Ah, right, I misread.  Honestly, I'm not sure, but I would maybe lean towards the way it's being done currently.  The extra query should be cheap-ish since it's only working with a single, indexed field.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -77,12 +77,39 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
...
+        )
+        content_ids_to_delete = premigrated_content_ids - mongo_content_ids
```
ipanova, line 99: the only suggestion i would have is to check whether this set is not empty before performing the query

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -77,12 +77,39 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
...
+            pulp2_id__in=content_ids_to_delete
+        ).delete()
+
     batch_size = 100
     pulp2content = []
     pulp2mutatedcontent = []
     content_type = content_model.pulp2.TYPE_ID
     set_pulp2_repo = content_model.pulp_2to3_detail.set_pulp2_repo
 
+    delete_removed_pulp2_content(content_model)
```
ipanova, line 111: you can probably pass content type from line 108

pulp_2to3_migration/app/pre_migration.py
goosemania: I expect it to find nothing veery quickly but I can add the check

pulp_2to3_migration/app/pre_migration.py
goosemania: I need the model anyway to perform the query in mongo.

pulp_2to3_migration/tests/functional/test_rpm_rerun.py
```diff
@@ -343,6 +349,21 @@ def test_rpm_distributors_only_added_or_with_changed_config(self):
 
         self.assertEqual(new_distribution_count, self.repo_info['new_distributions'])
 
+    def test_old_premigrated_content_removed(self):
+        """
+        Test that premigrated content is removed if the corresponding Pulp 2 content is removed.
```
ipanova, line 354: test that premigrated orphaned content

ipanova: not sure whether it worth pointing this out but at least it would be accurate based on the snapshots

ipanova (approved):

pulp_2to3_migration/app/pre_migration.py
ipanova: i know, i just personally dislike defining another variable when you already have it defined, matter of taste i guess

pulp_2to3_migration/tests/functional/test_rpm_rerun.py
goosemania: There is no such thing as orphaned premigrated content. Premigrated content is in Pulp2Content, it can't be orphaned, imo. Or we are introducing new terms :)
I initially wrote this comment with the word `orphaned` somewhere and it was very confusing, so I just wrote that the Pulp 2 content is removed.

ipanova: orphaned in sense that when it was premigrated it did not belong to any of the pulp2 repos

pulp_2to3_migration/app/pre_migration.py
goosemania: pushed

pulp_2to3_migration/tests/functional/test_rpm_rerun.py
goosemania: It might have belonged to some, maybe it became orphaned and was removed in between on migration runs.
The key is that it was pre-migrated but not migrated. And on the next run it dissappeared from pulp2 completely.

PR #344, opened 2021-03-30
Add missing indices

goosemania:
It should potentially improve performance of some queries a bit.
No benchmarking is done, just those fields are used in various queries.

closes #6742
https://pulp.plan.io/issues/6742


goosemania:
The only fields which are used in queries which I didn't add are the boolean ones, I do not think it makes sense to index those but let me know if I'm off here.

ipanova (approved):

pulp_2to3_migration/app/models/content.py
```diff
@@ -49,6 +49,7 @@ class Meta:
         ]
         indexes = [
             models.Index(fields=['pulp2_content_type_id']),
+            models.Index(fields=['pulp2_last_updated']),
```
goosemania, line 52: Usage:
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/pre_migration.py#L116
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/pre_migration.py#L155
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/pre_migration.py#L177

pulp_2to3_migration/app/models/repository.py
```diff
@@ -65,6 +65,9 @@ class Pulp2Repository(BaseModel):
     class Meta:
         indexes = [
             models.Index(fields=['pulp2_repo_type']),
+            models.Index(fields=['pulp2_last_unit_added']),
+            models.Index(fields=['pulp2_last_unit_removed']),
+            models.Index(fields=['pulp2_repo_id']),
```
goosemania, line 70: Usage:
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/migration.py#L84
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/migration.py#L99
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/migration.py#L184
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/models/base.py#L497

pulp_2to3_migration/app/models/repository.py
```diff
@@ -178,4 +182,6 @@ class Meta:
         unique_together = ('pulp2_object_id',)
         indexes = [
             models.Index(fields=['pulp2_type_id']),
+            models.Index(fields=['pulp2_last_updated']),
+            models.Index(fields=['pulp2_repo_id']),
```
goosemania, line 186: Usage:
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/migration.py#L212

goosemania: I found very similar for the importer, should I add the index to the importer as well?  
https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/migration.py#L165

dralley: It's up to you - my personal view is that certain indexes for things like Repositories and Importers aren't *that* important comparatively since we are dealing with a much smaller number of them (vs. content, artifacts, LCEs)

But it proabably doesn't hurt.

dralley: @goosemania If there are any queries that especially need to be fast, and they use multiple fields at once, we might want to add multi-column indexes (as opposed to indexes for multiple columns).

But it's easier for Postgresql to combine multiple indexes than it is to use a multi-column index for a single column search, so this is fine for now I think.  It's still better than having none at all as far as I can tell even if it's not optimally efficient.

https://stackoverflow.com/questions/39297221/multiple-indexes-vs-single-index-on-multiple-columns-in-postgresql
https://www.postgresql.org/docs/9.2/indexes-bitmap-scans.html
https://docs.djangoproject.com/en/3.1/ref/models/indexes/#condition

PR #345, opened 2021-04-05
Updating CI

fao89:
[noissue]


template_config.yml
```diff
@@ -1,8 +1,8 @@
 # This config represents the latest values used when running the plugin-template. Any settings that
 # were not present before running plugin-template have been added with their default values.
 
-additional_plugins:
-- branch: master
+additional_repos:
+- branch: "1.6"
   name: pulp_file
```
fao89, line 6: I had to pin pulp_file, because pulp_file@master uses `Distribution`, and the other plugins use `BaseDistribution`. The CI was breaking because of https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/models/repository.py#L173

Ref: https://github.com/pulp/pulp_file/pull/484

fao89: ^ FYI: @goosemania @ggainey @dralley @ipanova

goosemania: Good find!

It will be a pain to change that :( 
I foresee backports...backports...

goosemania: I think we'd need to pin everything until we fix it here.

goosemania (approved):
Thanks for catching the problem!
I'll update other plugin versions with a follow-up PR. The are not causing failures only because some plugin PRs have not been merged yet.

PR #346, opened 2021-04-07
Pin plugin versions until the Distribution change is supported.

goosemania:
[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -2,13 +2,13 @@
 # were not present before running plugin-template have been added with their default values.
 
 additional_repos:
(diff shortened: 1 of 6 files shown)


template_config.yml
```diff
@@ -2,13 +2,13 @@
 # were not present before running plugin-template have been added with their default values.
 
 additional_repos:
-- branch: "1.6"
+- branch: '1.6'
```
dralley, line 5: It's super confusing that the branch name goes before the name : /

ipanova (approved):

PR #347, opened 2021-04-07
Add CONTENT_PREMIGRATION_BATCH_SIZE configuration option

goosemania:
closes #8470
https://pulp.plan.io/issues/8470


pulp_2to3_migration/app/pre_migration.py
```diff
@@ -103,7 +104,7 @@ def delete_removed_pulp2_content(content_model):
                 pulp2_id__in=content_ids_to_delete
             ).delete()
 
-    batch_size = 100
+    batch_size = settings.CONTENT_PREMIGRATION_BATCH_SIZE or DEFAULT_BATCH_SIZE
```
ipanova, line 107: settings.get(CONTENT_PREMIGRATION_BATCH_SIZE, DEFAULT_BATCH_SIZE)

docs/configuration.rst
```diff
@@ -77,6 +77,17 @@ the migration is done and remove those two checksum types from the allowed list.
...
+4. Configure `CONTENT_PREMIGRATION_BATCH_SIZE` if needed.
+If your Pulp 2 setup is large and your system is relatively slow in terms of I/O (e.g. you have
+HDD), consider adjusting the batch size for content premigration. The default is 1000.
+
+It is recommended to decrease it gradually, approximately by half each time. On a very slow
+system and with many errata content to migrate, it might need to go as low as 50. Decreasing the
+value increases the time of content migration from pulp 2 to pulp 3. It's noticeable for a large
+setups only.
+
+The main sign that `CONTENT_PREMIGRATION_BATCH_SIZE` needs to go down is the ``pymongo.errors.CursorNotFound: Cursor not found`` errors in logs.
```
ipanova, line 89: :+1:

pulp_2to3_migration/app/pre_migration.py
goosemania: I didn't use get on purpose.
I expect the option always to be present since it's a part of the plugin default settings. My concern is that it can be set to None or 0. And in this case `get` doesn't help.

goosemania: Maybe stronger validation is needed to ensure that it's a positive integer.

ipanova: imo that's an overkill :D i can't recall us doing such validation on settings. I think the way you have it now is good enough

ipanova (approved):

PR #353, opened 2021-04-22
Fix corrupted content skipping for deb content

quba42:
Could be the fix for: https://pulp.plan.io/issues/8562 (I am not yet sure...)

### DIFF
--- a/pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
+++ b/pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
@@ -449,6 +449,11 @@ class Pulp2DebComponentPackage(Pulp2to3Content):
             relative_path=self.package_relative_path,
             sha256=self.package_sha256,
         ).first()
+        if not package:
+            # Perhaps the package was never created because the file was corrupt!
+            # We simply don't create the PackageReleaseComponent (just how the Package
+            # itself was not created) and hope for the best.
+            return (None, None)
         pulp3_package_release_component = pulp3_models.PackageReleaseComponent(
             release_component=release_component,
             package=package,
--- a/CHANGES/8612.bugfix
+++ b/CHANGES/8612.bugfix
@@ -1,0 +1,1 @@
+Fixed a bug in the deb pipeline that was preventing successfull skipping of corrupted content for migrations with ``skip_corrupted=True``.


pulp_2to3_migration/app/plugin/deb/pulp_2to3_models.py
```diff
@@ -449,6 +449,10 @@ def create_pulp3_content(self):
             relative_path=self.package_relative_path,
             sha256=self.package_sha256,
         ).first()
+        if not package:
+            # Perhaps the package was never created because the file was corrupt!
+            # We simply don't create the PackageReleaseComponent and hope for the best.
+            return ???
```
quba42, line 455: What can i return to make sure that the unit is eventually marked as migrated in the `Pulp2Content` table, but simply does not create any corresponding Pulp 3 unit?

ipanova: You can `return None, None` https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/plugin/content.py#L285

If it is not critical to create PackageReleaseComponent in case the corresponding package has not been migrated/save then this approach should be fine.

quba42: PackageReleaseComponent simply associates a Package with a ReleaseComponent (for the sake of generating repo metadata). If the package is not migrated/created/saved, then this association does not make any sense and should not happen either. The resulting migrated repository simply won't contain the corrupt package.

Of course, I should go ahead and test this by deliberately corrupting a package before migrating.

quba42:
I just ran a migration (without this change) after intentionally corrupting a package, which resulted in the following error: 
```
pulp_deb.app.models.content.PackageReleaseComponent.package.RelatedObjectDoesNotExist: PackageReleaseComponent has no package.
```
With the change the migration successfully completed instead.

ipanova:
> I just ran a migration (without this change) after intentionally corrupting a package, which resulted in the following error:
> 
> ```
> pulp_deb.app.models.content.PackageReleaseComponent.package.RelatedObjectDoesNotExist: PackageReleaseComponent has no package.
> ```
> 
> With the change the migration successfully completed instead.

this error tells that unfortunately the root cause for https://pulp.plan.io/issues/8562 is different :/

ipanova (approved):

goosemania (approved):

PR #354, opened 2021-04-23
Update total counter before migration of a content type starts.

goosemania:
Also unified some names in the progress report.

closes #8621
https://pulp.plan.io/issues/8621

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -233,8 +233,7 @@ class ContentMigrationFirstStage(Stage):
             pulp_2to3_detail_qs = pulp_2to3_detail_qs.order_by('repo_id')
 
         with ProgressReport(
-            message='Migrating {} content to Pulp 3 {}'.format(self.migrator.pulp2_plugin,
-                                                               content_type),
+            message='Migrating {} content to Pulp 3'.format(content_type),
             code='migrating.{}.content'.format(self.migrator.pulp2_plugin),
             total=pulp_2to3_detail_qs.count()
         ) as pb:
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -134,13 +134,13 @@ def pre_migrate_content_type(content_model, mutable_type, lazy_type, premigrate_
         total=total_content))
 
     pulp2content_pb = ProgressReport(
-        message='Pre-migrating Pulp 2 {} content (general info)'.format(content_type.upper()),
+        message='Pre-migrating Pulp 2 {} content (general info)'.format(content_type),
         code='premigrating.content.general',
         total=total_content,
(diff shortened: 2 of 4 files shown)


pulp_2to3_migration/app/plugin/content.py
```diff
@@ -233,8 +233,7 @@ def get_remote_by_importer_id(importer_id):
             pulp_2to3_detail_qs = pulp_2to3_detail_qs.order_by('repo_id')
 
         with ProgressReport(
-            message='Migrating {} content to Pulp 3 {}'.format(self.migrator.pulp2_plugin,
-                                                               content_type),
+            message='Migrating {} content to Pulp 3'.format(content_type),
```
goosemania, line 236: Was `Migrating iso content to Pulp 3 iso`, `Migrating rpm content to Pulp 3 erratum`.
Now ` Migrating iso content to Pulp 3`, `Migrating erratum content to Pulp 3 `.

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -134,13 +134,13 @@ def delete_removed_pulp2_content(content_model):
         total=total_content))
 
     pulp2content_pb = ProgressReport(
-        message='Pre-migrating Pulp 2 {} content (general info)'.format(content_type.upper()),
+        message='Pre-migrating Pulp 2 {} content (general info)'.format(content_type),
```
ggainey, line 137: Instead of dropping upper() here, maybe leave these as they are and 'upper' the message at content:236? I found the UPPER words easier to pick out.

That may just be a me-thing, however, I'm not tied to it!

ggainey (changes requested):

pulp_2to3_migration/app/pre_migration.py
goosemania: That's funny. The are/were upper() exactly for the readability but I thought it's just my thing and someone was asking to make it lowercase, maybe it was katello :)

goosemania: https://pulp.plan.io/issues/7732

ggainey: Ha! Well, I find the caps make it easier to see what's going on - without them, the response becomes a paragraph, and it's harder for the eye to pick out what's going on. I def prefer "ease of finding the critical info" over "pretty". If I were submitting the PR, I'd leave the caps, but I certainly won't block merging if you decide to go w/ jsherril's suggestion.

goosemania: I think I'd go with lowercase option, just to be on the safe side. Different people can read/interpret caps as something other than just the intention to make it readable.

ggainey: Sounds good to me :)

ggainey (approved):

PR #355, opened 2021-04-24
Change task spawning to use the new dispatch() function

dralley:
[noissue]


dralley:
This can wait until we're ready to bump the pulpcore version floor.

PR #356, opened 2021-04-24
Migrate relationship between repo and importer to Pulp 3

dralley:
Populate the "remote" field on the Pulp 3 repository model with the
information stored in the Pulp2Repositories table, so that it can
survive post-migration.

closes: #8626
https://pulp.plan.io/issues/8626

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -425,15 +425,18 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
                 # exclude the resolved path from further search
                 paths.remove(bad_path)
 
+    pulp3_repo = pulp2_repo.pulp3_repository
     if pulp3_remote:
         pulp2_repo.pulp3_repository_remote = pulp3_remote
         pulp2_repo.save()
+
+        pulp3_repo.remote = pulp3_remote
+        pulp3_repo.save()
 
     if pulp2_repo.is_migrated:
         progress_rv.update(total=F('total') - 1)
         return
 
-    pulp3_repo = pulp2_repo.pulp3_repository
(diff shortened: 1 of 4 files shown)


pulp_2to3_migration/app/migration.py
```diff
@@ -321,15 +321,18 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
         pulp3_remote(remote): a pulp3 remote
     """
 
+    pulp3_repo = pulp2_repo.pulp3_repository
     if pulp3_remote:
         pulp2_repo.pulp3_repository_remote = pulp3_remote
         pulp2_repo.save()
 
+        pulp3_repo.remote = pulp3_remote
+        pulp3_repo.save()
```
quba42, line 330: I am not sure this would work for all plugins. In Pulp 3 each plugin needs to implement this association for it to be available. For `pulp_deb` I only happened to stumble across this recently (See https://github.com/pulp/pulp_deb/pull/261) but it is not yet released.

dralley: @quba42 This appears to be released now, is that correct?

goosemania: @quba42 , this is planned to go into 0.12 release which will be compatible with pulpcore 3.12-3.14.
This release will no longer compatible with Katello. They are staying on 0.11.z and won't go beyond that.

goosemania (approved):

pulp_2to3_migration/app/migration.py
quba42: The pulp_deb change was released with version 2.12.0 which is compatible with `pulpcore>=3.12,<3.14`, so that should line up nicely.

PR #357, opened 2021-04-27
handle null deb importer attributes

jlsherrill:
fixes #8613

### DIFF
--- a/pulp_2to3_migration/app/plugin/deb/repository.py
+++ b/pulp_2to3_migration/app/plugin/deb/repository.py
@@ -34,9 +34,12 @@ class DebImporter(Pulp2to3Importer):
         pulp2_config = pulp2importer.pulp2_config
         base_config, name = cls.parse_base_config(pulp2importer, pulp2_config)
         base_config['name'] = name
-        base_config['distributions'] = pulp2_config.get('releases').replace(',', ' ')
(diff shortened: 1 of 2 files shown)


jlsherrill:
Note, i haven't actually tested this yet :)

pulp_2to3_migration/app/plugin/deb/repository.py
```diff
@@ -34,9 +34,12 @@ def migrate_to_pulp3(cls, pulp2importer):
         pulp2_config = pulp2importer.pulp2_config
         base_config, name = cls.parse_base_config(pulp2importer, pulp2_config)
         base_config['name'] = name
-        base_config['distributions'] = pulp2_config.get('releases').replace(',', ' ')
-        base_config['components'] = pulp2_config.get('components').replace(',', ' ')
-        base_config['architectures'] = pulp2_config.get('architectures').replace(',', ' ')
+        if pulp2_config.get('releases') != None:
+            base_config['distributions'] = pulp2_config.get('releases').replace(',', ' ')
```
jlsherrill, line 38: technically distributions seems to be required (at least according to the rest docs for pulp3), but it wasn't required in pulp2.  I'm not sure what we can do here, but i think the error that a blank distributions would throw might be better than this 'Nonetype object has no attribute replace'.  But i'm fine removing these two lines, or open to other suggestions

PR #360, opened 2021-04-29
Do not migrate unsupported metadata.

goosemania:
Exclude from premigration any data type which ends with _zck or _xz in repomd.xml.

closes #8400
https://pulp.plan.io/issues/8400

### DIFF
--- a/CHANGES/8400.bugfix
+++ b/CHANGES/8400.bugfix
@@ -1,0 +1,1 @@
+Stopped migrating unsupported metadata, like .zck, which could have been imported into some old Pulp 2 version.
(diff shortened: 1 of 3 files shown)


pulp_2to3_migration/app/plugin/rpm/yum_repo_metadata.py
```diff
@@ -0,0 +1,13 @@
+from mongoengine.queryset.visitor import Q as mongo_Q
```
ipanova, line 1: my only suggestion would be to move this to plugins/rpm/utils.py instead

ipanova (approved):

pulp_2to3_migration/app/plugin/rpm/yum_repo_metadata.py
goosemania: @ipanova , did you mean xml_utils.py? or do you suggest to rename this file to utils.py?

ipanova: My suggestions was to rename the file to utils.py

PR #361, opened 2021-04-30
Handle overlapping paths which come from pulp2 repos

goosemania:
closes #8582
https://pulp.plan.io/issues/8582

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -1,14 +1,19 @@
 import logging
 
 from django.db.models import F, Q
+from gettext import gettext as _
+
+from pygtrie import StringTrie
 
 from pulpcore.plugin.models import (
     Content,
+    ContentArtifact,
     CreatedResource,
     ProgressReport,
     Repository,
     TaskGroup,
 )
+
 from pulpcore.plugin.tasking import enqueue_with_reservation
 
 from pulp_2to3_migration.app.models import (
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +321,67 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
         pulp2_repo(Pulp2Repository): a pre-migrated repository to create a repo version for
         pulp3_remote(remote): a pulp3 remote
     """
+    def detect_path_overlap(paths):
+        """
+        Check for valid POSIX paths (ie ones that aren't duplicated and don't overlap).
+
+        Overlapping paths are where one path terminates inside another (e.g. a/b and a/b/c).
+
+        NOTE: The logic is copied from pulpcore.app.files.validate_file_paths().
```
goosemania, line 330: Using directly from pulpcore is painful because I would need to extract paths from the exception messages.

pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +321,67 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
...
+
+        This function returns the first dupe or overlap it detects. We use a trie (or
+        prefix tree) to keep track of which paths we've already seen.
+
+        Args:
+            paths (iterable of str): An iterable of strings each representing a relative path
+
+        Returns:
+            str: a path which overlaps another
```
goosemania, line 339: I chose an easy way, just deal with the path which is conflicting and not analyse the list of conflicts and try to come up with some other logic. What do you think? See the original implementation https://github.com/pulp/pulpcore/blob/master/pulpcore/app/files.py#L113

pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +321,67 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
...
+        iteration = 0
+        while True and iteration <= max_conflicts:
+            bad_path = detect_path_overlap(paths)
+            if not bad_path:
+                # no path overlaps, we are good
+                break
+
+            # find content by its relative path
+            # if duplicated - find pulp2content and check pulp2 timestamps, remove older one.
+            # otherwise remove the one found from the version
```
goosemania, line 383: I'm currently looking into how to implement this logic, and it seems somewhat painful.
Is it overkill to try to find the latest one? Should we remove just the first one?
FWIW, pulp 2 publishes the first met one I think, at least out of 2 content units with the same relative path, it picked and published the older one :/

Any suggestions are welcome, for this logic or in general.

dralley: Maybe you could do something like the "retain old count" - partition by `relative_path`, use an expression to get the first value.

https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/package.py#L32-L54
https://docs.djangoproject.com/en/3.2/ref/models/database-functions/#firstvalue

dralley: Disclaimer: I have no idea if it would work exactly like that, I don't know much about window functions

goosemania: I ended up doing just sorting and removing everything but not the most recent one from the repo version.
Thanks for the suggestions.

pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +324,89 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
...
+            ).order_by('-content__pulp2content__pulp2_last_updated')
+
+            conflict_count = cas_with_conflicts.count()
+            if conflict_count > 1:
+                # There are duplicated paths and we need to keep the newest content in the version.
+                # The query result is ordered desc by time so we remove all but the first content.
+                content_to_remove = [ca.content.pk for ca in cas_with_conflicts[1:]]
+                version.remove_content(Content.objects.filter(pk__in=content_to_remove))
+                # exclude the duplicated paths from further search
+                for j in range(conflict_count - 1):
```
ggainey, line 402: Can we get an INFO log here, "Removing {conflict_count -1} instance of {bad_path}" here, so we have an answer when someone says "this repo had N rpms in Pulp2, and only N-M in pulp3 - what happened?!?"

pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +324,89 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
...
+                    paths.remove(bad_path)
+            else:
+                # It's not a duplicated path but it overlaps with some other in the version,
+                # it should be removed from the version to resolve the conflict.
+                version.remove_content(cas_with_conflicts[0].content)
```
ggainey, line 407: Same here

ggainey (changes requested):
Looks good - just looking for some reporting when we have to change things, to make it easier to respond to user-questions post-migration.

pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +324,89 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
...
+            if prefixes:
+                # overlap where path is 'a/b/c' and trie has 'a/b'
+                return path
+
+            # if there are no overlaps, add it to our trie and continue
+            path_trie[path] = True
+
+    def resolve_path_overlap(version):
+        """
+        Remove content which path overlaps some other.
```
dralley, line 365: minor typo:  s/which/for which

pulp_2to3_migration/app/migration.py
```diff
@@ -320,6 +324,89 @@ def create_repo_version(progress_rv, pulp2_repo, pulp3_remote=None):
...
+        paths = ContentArtifact.objects.filter(content__pk__in=version.content).values_list(
+            "relative_path", flat=True
+        )
+        paths = list(paths)
+        max_conflicts = version.content.count() - 1
+        for i in range(max_conflicts):
+            bad_path = detect_path_overlap(paths)
+            if not bad_path:
+                # no path overlaps, we are good
+                break
```
dralley, line 388: Should this be `continue`?  Do we need to keep checking other paths?

dralley: Ah, maybe not.  Because we're passing the whole list in here.  The loop confused me, but I guess it's only being used for resolution of overlaps?

goosemania: I could have had a while loop here, but just to be on the safe side, in case things go south, we are not stuck in the infinite loop.
It will loop only if there are any conflicts.

pulp_2to3_migration/app/migration.py
ipanova: my suggestion would be to not pollute logging with N messages, instead how about notifying the user once about the fact that there is content that has some path overlap issue. I doubt user needs to know that path exactly is being solved.
```
for i in range(max_conflicts):
    ...
else:
   log.info('Some content had  overlapping paths and was removed')
``

Thoughts?

ipanova: Or maybe not. It seems like The for loop will will always hit 'break' sooner or later.

ipanova: we can always compare the len of paths before entering the loop and after exiting it;)

goosemania: Yeah, this code executes only when there are conflicts. For the majority of repo verisons, nothing will be printed.

ggainey (approved):
lgtm, thanks!

ipanova (approved):

PR #371, opened 2021-05-18
Use dispatch instead of enqueue_with_reservation

mdellweg:
Bump pulpcore dependency to 3.12 accordingly.

[noissue]

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -14,7 +14,7 @@ from pulpcore.plugin.models import (
     TaskGroup,
 )
 
-from pulpcore.plugin.tasking import enqueue_with_reservation
+from pulpcore.plugin.tasking import dispatch
 
 from pulp_2to3_migration.app.models import (
     Pulp2Content,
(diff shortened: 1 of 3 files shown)


ipanova:
you need to str(pk) like it was done here https://github.com/pulp/pulp-2to3-migration/pull/355/files

mdellweg:
> you need to str(pk) like it was done here https://github.com/pulp/pulp-2to3-migration/pull/355/files

Actually, we decided that we want to take this inconvenience off of our plugin writers:
https://github.com/pulp/pulpcore/commit/b1b9b2cf7ada47ba43f656983db5b690100b3165

ipanova:
> > you need to str(pk) like it was done here https://github.com/pulp/pulp-2to3-migration/pull/355/files
> 
> Actually, we decided that we want to take this inconvenience off of our plugin writers:
> [pulp/pulpcore@b1b9b2c](https://github.com/pulp/pulpcore/commit/b1b9b2cf7ada47ba43f656983db5b690100b3165)

yes, i am aware of that, but won't it be available from 3.13?

mdellweg:
> > > you need to str(pk) like it was done here https://github.com/pulp/pulp-2to3-migration/pull/355/files
> > 
> > 
> > Actually, we decided that we want to take this inconvenience off of our plugin writers:
> > [pulp/pulpcore@b1b9b2c](https://github.com/pulp/pulpcore/commit/b1b9b2cf7ada47ba43f656983db5b690100b3165)
> 
> yes, i am aware of that, but won't it be available from 3.13?

It has and will always work. It's "just" erroneously deprecated in 3.12.

dralley:
I already made this PR (https://github.com/pulp/pulp-2to3-migration/pull/355) the reason I closed it is that we can't do this until the migration plugin is ready to bump the pulpcore compatibility version floor, and I think that wasn't going to be for a while IIRC.

PR #374, opened 2021-05-27
Remove relations only for plugins specified in the plan

goosemania:
closes #8686
https://pulp.plan.io/issues/8686

### DIFF
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -716,8 +716,9 @@ def handle_outdated_resources(plan):
     Args:
         plan(MigrationPlan): A Migration Plan
     """
-    RepoSetup.mark_changed_relations()
-    RepoSetup.finalize()
+    plugins = plan.get_plugins()
+    RepoSetup.mark_changed_relations(plugins)
+    RepoSetup.finalize(plugins)
 
     for plugin_plan in plan.get_plugin_plans():
         inplan_repos = plugin_plan.get_repositories()
--- a/pulp_2to3_migration/tests/functional/test_multiple_plugins.py
+++ b/pulp_2to3_migration/tests/functional/test_multiple_plugins.py
@@ -1,0 +1,145 @@
+import time
+import unittest
+
+from pulpcore.client.pulpcore import Configuration
+from pulpcore.client.pulp_file import (
+    ApiClient as FileApiClient,
+    ContentFilesApi,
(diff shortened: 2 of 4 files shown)


goosemania:
tests pass locally, I'm trying to figure out what's wrong with ci

pulp_2to3_migration/app/pre_migration.py
```diff
@@ -716,8 +716,9 @@ def handle_outdated_resources(plan):
     Args:
         plan(MigrationPlan): A Migration Plan
     """
-    RepoSetup.mark_changed_relations()
-    RepoSetup.finalize()
+    plugins = plan.get_plugins()
```
ggainey, line 719: :+1:

ggainey (approved):
Great catch - looks good to me (assuming we can get the tests to pass, of course :) )

pulp_2to3_migration/tests/functional/test_multiple_plugins.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+    def test_rpm_iso_migration_sequential(self):
+        """Test migrating RPM plugin and ISO plugin in two separate runs"""
+        set_pulp2_snapshot(name='rpm_base_4repos')
+        self.run_migration(RPM_SIMPLE_PLAN)
+
+        set_pulp2_snapshot(name='file_base_4repos')
+        self.run_migration(FILE_SIMPLE_PLAN)
+
+        rpm_repo_count = 3
```
ipanova, line 138: took me a while to understand why the count is 3. in that snapshot are 5 repos, from which 2 are empty. not sure what to suggest to make this more explicit?..and why this is called rpm_base_4repos while in the snapshot there are 5? https://github.com/pulp/pulp-2to3-migration-test-fixtures/blob/master/rpm_base_4repos/description.md

ipanova (approved):

pulp_2to3_migration/tests/functional/test_multiple_plugins.py
goosemania: It's very confusing I agree. 
I noticed the snapshot naming while prepping this pr as well. I think originally it was 4 and then we tweaked it and never changed the name. I'll try to redo it at some point to avoid confusion around 4 in name vs 5 in reality.
Simple plan does not migrate empty repos, so I can provide complex plan or leave a comment that empty repos are not migrated using simple plan.

PR #377, opened 2021-06-04
Fix migration of on_demand distribution trees which do not have images.

goosemania:
closes #8817
https://pulp.plan.io/issues/8817


dralley (approved):
Looks reasonable based on the comments and the description in the issue.

PR #378, opened 2021-06-07
Distribution changes

ggainey:


pulp_2to3_migration/app/models/repository.py
```diff
@@ -175,7 +175,7 @@ class Pulp2Distributor(BaseModel):
 
     # due to base_path overlap restriction, a distribution can't correspond to multiple pulp 2
     # distributors, thus one-to-one relationship.
-    pulp3_distribution = models.OneToOneField(BaseDistribution,
+    pulp3_distribution = models.OneToOneField(Distribution,
```
dralley, line 178: We will need to migrate/rewrite these columns - but the UUID key itself within should remain the same.

ggainey: ...I failed to add the migration to the commit. Sigh. One sec, I'll add and repush.

dralley: Remember to bump the required version of pulpcore to >=3.12 as well

pulp_2to3_migration/app/migrations/0029_distribution_change.py
```diff
@@ -0,0 +1,19 @@
...
+
+    dependencies = [
+        ('pulp_2to3_migration', '0028_create_missing_indices'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='pulp2distributor',
+            name='pulp3_distribution',
+            field=models.OneToOneField(null=True, on_delete=django.db.models.deletion.SET_NULL, to='core.Distribution'),
```
dralley, line 17: This keeps the data intact, yes?

dralley (approved):
Looks good to me.

My only concern is that this will lock us out of making migration changes to the 3.7 compatible branch.  Which is completely expected, but if we end up needing to do so, it will be a bad time.

On the other hand, we have users that are blocked on this currently.

ggainey:
> Looks good to me.
> 
> My only concern is that this will lock us out of making migration changes to the 3.7 compatible branch. Which is completely expected, but if we end up needing to do so, it will be a bad time.
> 
> On the other hand, we have users that are blocked on this currently.

The concern is valid - but as soon as we made backwards-incompatible changes to pulpcore, we were going to get here. The only other choice is to freeze 2to3 at core-3.11. As you point out, that means that a pulp-2.21.5 user that wants to use pulp3 **today**, has to start with back-level versions of core and the plugins they care about, migrate, and **then** move to Today's Pulp3. Which is equally painful.

The problem of "you can't backport fixes that require a migration" is the thorny, smelly elephant in the room; this particular sequence just makes it (even more) glaringly obvious.

goosemania:
@ggainey , could you share the scenario(s) you tested? I'm trying to figure out how the Distributors are getting removed and on the migration re-run the base_path overlap is not happening.

ggainey:
- set up pulp2-nightly-pulp3-source-centos7 as core-3.10 and friends (file/rpm/container/deb/2to3)
- pulp2 sync rpm-with-modules
- 2to3-migrate all rpm content
- update core et al to master, and this branch in 2to3
- pulpcore-manager migrate
- check table in psql, note distribution is null
- prestart
- run 2to3-migration again
- look at tables in psql, note that distribution has been migrated

ggainey:
> @ggainey , could you share the scenario(s) you tested? I'm trying to figure out how the Distributors are getting removed and on the migration re-run the base_path overlap is not happening.

I reset and re-ran my scenario this morning and I notice that, on re-running the migration, the distribution is recreated, but we do **not** appear to republish.

dralley:
The migration plugin doesn't have hard requirements on any plugins (they can be present or not-present) but if they *are* present, we will now have a requirement on which versions are supported.

Do we need to throw a runtime error if the user has an unsupported plugin installed?

goosemania:
@ggainey , I tested RPM, Docker an ISO plugins, migrating them all together.
It works correctly only for RPM, and looking at your code I do not understand why.

But here are my observations:
 - before new migration is applied, all repos have a distribution in pulp3
 - after the django migration is applied,  all distributions are unlinked in pulp2distributor table (which is correct) but only rpm ones are marked as is_migrated = False (which is incorrect).
 - after running migration from pulp2 to pulp3, the distribution are restored only for RPM repos (which is incorrect but saomewhat expected if is_migrated stayed =True).

To test file and docker on the migration box, you can use `populate_pulp2_docker  populate_pulp2_iso` commands available in the pulp venv of the box. And jfyi, docker repos do not have publications in pulp3, just a distribution.

ggainey:
@goosemania I see what you're seeing. In my testing, the filter() works correctly, and the update() updates the pulp_distribution to null, but **only** the rpm-dist gets is_migrated set to 'f'.

Issuing an update 'by hand' in pulpcore-manager shell does the right thing.

I can't find anything in django-docs about update() that would suggest this behavior. Investigation continues.

goosemania:
> > @ggainey , could you share the scenario(s) you tested? I'm trying to figure out how the Distributors are getting removed and on the migration re-run the base_path overlap is not happening.
> 
> I reset and re-ran my scenario this morning and I notice that, on re-running the migration, the distribution is recreated, but we do **not** appear to republish.

So I think I figured that out. And there are some problems. Let me share what I found and then we can decide what to do.
1. Publications/Distributions are not removed because [we are checking it based on the pulp2repository information](https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/pre_migration.py#L772).
2. Publications and distributions are magically re-linked, because
-  [if publication exists it will be used](https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/plugin/rpm/repository.py#L69) 
- nothing substantial changed for a distributor and [it uses update_or_create](https://github.com/pulp/pulp-2to3-migration/blob/master/pulp_2to3_migration/app/plugin/rpm/repository.py#L91)
3. The problems come when
- you change relative url in pulp 2 (`pulp-admin rpm repo update --repo-id <your_repo_id> --relative-url <new_relative_url>`). You'll see `django.db.utils.IntegrityError: duplicate key value violates unique constraint "core_distribution_name_key"`
- after doing ^, you also create a new repo/distributor with the old relative url from ^. You'll see base path overlap issue, because pulp3 distributor is not removed. (I haven't tested this one but that's my deduction from reading the code.)

Ways to mitigate the issues above:
 - in django migration, set is_migrated=False for repositories as well, it will cause all publications/distributions to be recreated.
 - at cleaning up stage, remove all distributions for plugins in a migration plan which are not linked in pulp2distributor table. This saves time on republishing.
 - do as we have now and if users see any issues during migration plugin run, they should run reset first. But advise in the release notes to run pulp 2to3 migration before the upgrade to minimize chance to have pulp2 changes for distributors.

ggainey:
@goosemania So the failure case for "what we're doing right now" would be:

- be on core-3.11
- do a migration
- update repos on Pulp2
- upgrade to 3.13+
- re-migrate

Is that correct? If so - this feels like an edge-case that could be handled by documentation, as you suggest in bullet-3 above. I'd like the impact of this change to affect main-path migrations as little as possible, and the other ways to fix it make more work for every migrating instance. That's my thinking right now - will mull on it overnight and see if it feels the same tomorrow!

goosemania:
> @goosemania So the failure case for "what we're doing right now" would be:
> 
>     * be on core-3.11
> 
>     * do a migration
> 
>     * update repos on Pulp2
> 
>     * upgrade to 3.13+
> 
>     * re-migrate
> 
> 
> Is that correct? If so - this feels like an edge-case that could be handled by documentation, as you suggest in bullet-3 above. I'd like the impact of this change to affect main-path migrations as little as possible, and the other ways to fix it make more work for every migrating instance. That's my thinking right now - will mull on it overnight and see if it feels the same tomorrow!

You would be surprised how common that edge-case can be :) we hoped for it to be rare in the past.
At the same time, I agree to rely on the reset operation, not to introduce more problems with complicated logic for the semi-supported use cases.

goosemania:
I retested with the new changes and everything works now as expected, thanks!

goosemania (approved):

PR #380, opened 2021-06-10
Use checksum type of a package for publication if it's not configured.

goosemania:
This doesn't fix users in the middle of RPM plugin migration,
they would need to reset RPM migration and start it again.

closes #8725
https://pulp.plan.io/issues/8725

### DIFF
--- a/CHANGES/8725.bugfix
+++ b/CHANGES/8725.bugfix
@@ -1,0 +1,4 @@
+Ensure a checksum type of a package is used for publications when a checksum type was not explicitly configured in Pulp 2.
+
+If you plan to perform sync from the migrated Pulp 3 to a Pulp 2 instance, this fix is important, otherwise you can ignore it.
+If you've already started migration of the RPM plugin to Pulp 3, reset the migration for it and start again.
--- a/pulp_2to3_migration/app/plugin/rpm/repository.py
+++ b/pulp_2to3_migration/app/plugin/rpm/repository.py
@@ -71,8 +71,21 @@ class RpmDistributor(Pulp2to3Distributor):
             pulp2_checksum_type = pulp2_config.get('checksum_type')
             checksum_types = None
             if pulp2_checksum_type:
-                checksum_types = {'metadata': pulp2_checksum_type,
-                                  'package': pulp2_checksum_type}
+                checksum_types = {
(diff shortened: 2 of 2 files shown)


goosemania:
I'm trying to avoid a db migration.
The usual argument here is that people who are in the middle of the migration are in the minority and would need to run reset if they want to fix already migrated repos.
DB migration complicates backporting and can be a source of additional bugs.

CHANGES/8725.bugfix
```diff
@@ -0,0 +1,4 @@
+Ensure a checksum type of a package is used for publications when a checksum type was not explicitly configured in Pulp 2.
+
+If you plan to perform sync from the migrated Pulp 3 to a Pulp 2 instance, this fix is important, otherwise you can ignore it.
+If you've already started migration of the RPM plugin to Pulp 3, reset the migraiton for it and start again.
```
dralley, line 4: typo "migraiton"

goosemania: thanks, fixed

pulp_2to3_migration/app/plugin/rpm/repository.py
```diff
@@ -71,8 +71,21 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
...
+                package_qs = repo_version.content.filter(pulp_type='rpm.package')
+                if package_qs.count():
+                    pkg_checksum_type = package_qs.first().cast().checksum_type
+                    checksum_types = {
+                        'metadata': pkg_checksum_type,
+                        'package': pkg_checksum_type
+                    }
             sqlite = pulp2_config.get('generate_sqlite', False)
             try:
                 publish(repo_version.pk, checksum_types=checksum_types, sqlite_metadata=sqlite)
```
dralley, line 91: I think we can get rid of the "hack" now

dralley (approved):
LGTM

pulp_2to3_migration/app/plugin/rpm/repository.py
goosemania: This will go into 0.11, so I'd leave the hack.
We can remove it after we merge compatibility PRs for 3.13.

PR #384, opened 2021-06-14
Make the plugin compatible with pulpcore 3.14

goosemania:
New tasking system changes some models a bit.
We can remove the hack when/if we announce a compatibility with pulpcore 3.14+.

closes #8895
https://pulp.plan.io/issues/8895


goosemania:
Pulpcore is releasing a deprecation first, so this can be done later, the old way will be removed in 3.15

PR #386, opened 2021-06-15
Updates GHA workflows with latest from plugin_template

dkliban:
[noissue]


ggainey (approved):
template looks good, the rest is the result of plugin-template running

PR #387, opened 2021-06-15
Remove api.json

goosemania:
It is no longer used for API docs.
Add to .gitignore as well.

[noissue]


docs/_static/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Ignore everything in this directory
+*
+
+# Except
+!.gitignore
```
goosemania, line 5: Docs job uses _static directory to download api.json https://github.com/pulp/pulp-2to3-migration/pull/387/checks?check_run_id=2832199051#step:9:60.
Also the conf.py for docs claims that `_static` is the one to be used for static content.
And git doesn't see/keep empty dirs, so this is a workaround to preserve the empty dir in the git repo.

goosemania: Any other suggestions?

ggainey (approved):
lg2m

docs/_static/.gitignore
goosemania: Alternatively, there is a `.gitkeep` option, which I heard is not a good idea,  because this file contains git in it but has nothing to do with git itself.

daviddavis: I'm not super familiar with the docs process. How does the `_build` dir get created? Is `_static` only used for api.json? Would it be possible to just download `api.json` to `_build` or have the docs job create a `_static` directory.

Also this solution you have here is fine with me. So is having a .gitkeep file although I understand your concern.

daviddavis (approved):

PR #390, opened 2021-06-16
Copy settings.py from plugin into the settings.py.j2 template

dkliban:
[noissue]


goosemania (approved):

goosemania:
Alternatively, maybe we can specify in the env vars, either the file with the settings if it's available, or the missing configuration explicitly. Wdyt?

dkliban:
The env vars need to be set using the Containerfile. It was easier for me to modify the  settings.py template using the settings.py from the plugin than to modify the Containerfile template.

PR #396, opened 2021-06-18
Fix survey banner

gerrod3:
[noissue]

@goosemania I made my banner pr with the RTD theme in mind, so a couple changes was needed for the theme this plugin uses.

Screenshot from 2021-06-18 10-21-50


goosemania:
Ah, thanks for checking and fixing!

goosemania (approved):

PR #402, opened 2021-06-28
Add deb tests

m-bucher:
Trying to add some tests for debain-migration :crossed_fingers:

### DIFF
--- a/CHANGES/9185.misc
+++ b/CHANGES/9185.misc
@@ -1,0 +1,1 @@
+Add deb tests
--- a/pulp_2to3_migration/tests/functional/common_plans.py
+++ b/pulp_2to3_migration/tests/functional/common_plans.py
@@ -5,6 +5,55 @@ Migration plans which are used by many different tests.
 import json
 
 
+DEB_COMPLEX_PLAN = json.dumps({
+    "plugins": [{
+        "type": "deb",
+        "repositories": [
+            {
+                "name": "debian-empty",
+                "pulp2_importer_repository_id": "debian-empty",
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "debian-empty",  # content count: 0
+                        "pulp2_distributor_repository_ids": ["debian-empty"]
+                    }
+                ]
+            },
+            {
+                "name": "debian",
+                "pulp2_importer_repository_id": "debian",
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "debian",
+                        "pulp2_distributor_repository_ids": ["debian"]
+                    }
+                ]
+            },
+            {
+                "name": "debian-complex-dists",
+                "pulp2_importer_repository_id": "debian-complex-dists",
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "debian-complex-dists",
+                        "pulp2_distributor_repository_ids": ["debian-complex-dists"]
+                    }
+                ]
+            },
+            {
+                "name": "debian_update",
+                "pulp2_importer_repository_id": "debian_update",
+                "repository_versions": [
+                    {
+                        "pulp2_repository_id": "debian_update",
+                        "pulp2_distributor_repository_ids": ["debian_update"]
+                    }
+                ]
+            },
+        ],
+
+    }]
+})
+
 FILE_COMPLEX_PLAN = json.dumps({
     "plugins": [{
         "type": "iso",
@@ -116,6 +165,7 @@ RPM_COMPLEX_PLAN = json.dumps({
 
 FILE_SIMPLE_PLAN = json.dumps({"plugins": [{"type": "iso"}]})
 RPM_SIMPLE_PLAN = json.dumps({"plugins": [{"type": "rpm"}]})
+DEB_SIMPLE_PLAN = json.dumps({"plugins": [{"type": "deb"}]})
 
 FILE_DISTRIBUTOR_DIFF_PLAN = json.dumps({
     "plugins": [{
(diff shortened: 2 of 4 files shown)


goosemania:
please rebase, we fixed CI

m-bucher:
So one of the problems blocking the tests is that the `debian-empty`-repository is not migrated.
This could be expected behaviour.
But given that for RPM we have a similar test that runs through, I doubt it :thinking:

goosemania:
> So one of the problems blocking the tests is that the `debian-empty`-repository is not migrated.
> This could be expected behaviour.
> But given that for RPM we have a similar test that runs through, I doubt it

It is expected for an empty repo to be migrated if it's specified as a part of the complex plan (each repository is mentioned explicitly in the plan). Empty repos are not migrated as a part of a simple plan (no repos are mentioned explicitly, you request to migrate everything for a plugin).
Such inconsistency is due to the following technical reason.
When an empty repository is explicitly specified in a plan, it is listed as a part of a certain plugin and the migration plugin knows which plugin this repository belongs to. In case of a simple plan, migration plugin needs to figure out itself which plugin a repo belongs to. It does it by looking at the content, that's why it is not possible to do that for an empty repo with no content in it.

pulp_2to3_migration/app/plugin/deb/repository.py
```diff
@@ -78,7 +78,7 @@ def migrate_to_pulp3(cls, pulp2distributor, repo_version):
         pulp2_config = pulp2distributor.pulp2_config
         distribution_data = cls.parse_base_config(pulp2distributor, pulp2_config)
         base_path = pulp2_config.get('relative_url', pulp2distributor.pulp2_repo_id)
-        distribution_data['base_path'] = base_path.rstrip('/')
+        distribution_data['base_path'] = base_path.strip('/')
```
quba42, line 81: In effect this minor code change removes leading slashes from the `base_path` during migration.
My intuition is that leading slashes on the `base_path` do not make a difference either way, but that it is cleaner not to have them, so I am cautiously in favor of the code change.

goosemania: Do you need this to be backported to 0.11?
This change doesn't seem to relate to the PR title, it's a code change.
I can merge it as a part of test PR if not backport is needed. If you want to backport it to 0.11, please file an issue.
Actually if you prefer tests to be backported, please file an issue for them as well.

pulp_2to3_migration/tests/functional/scripts/set_pulp2.sh
```diff
@@ -3,11 +3,11 @@
 MONGO_HOST=$1
 ARCHIVE_NAME=$2
 
-[ ! -d "pulp-2to3-migration-test-fixtures" ] && git clone https://github.com/pulp/pulp-2to3-migration-test-fixtures
+[ ! -d "pulp-2to3-migration-test-fixtures" ] && git clone https://github.com/ATIX-AG/pulp-2to3-migration-test-fixtures.git -b add_debian_repos
```
goosemania, line 6: I merged the test fixtures for debian

goosemania:
Please ping me when it's ready to be merged and if you need any backports.

pulp_2to3_migration/app/plugin/deb/repository.py
quba42: As I understand it it was the simplest/only way to make all the tests turn green, and it makes sense as a non-critical code improvement (since leading slashes on a `base_path` should not hurt, but aren't meaningful either). The reason I originally used `rstrip` was copy and paste from other content types. It might make sense to backport the tests (along with this change) to 0.11, but it probably makes most sense to open a PR for that once this one has been merged (rather than using the usual cherry-pick process)?

m-bucher:
Extracted the code-change into https://github.com/pulp/pulp-2to3-migration/pull/419
@goosemania I will have to rebase this PR onto that one, once it is merged. Otherwise the tests will fail again :sweat_smile:

pulp_2to3_migration/app/plugin/deb/repository.py
goosemania: Both ways work for me.
I would cherry-pick anything needed and that's it but if you prefer to open separate PRs for each branch, feel free to do so.
It won't be acceptable for any change but because the change is small and the rest are tests, we can do your preferred way this time.

m-bucher: I opened a Backport-ticket for the code change (https://pulp.plan.io/issues/9188)
For the tests, we will open a separate PR so we can react to failing tests (if they occur).

quba42 (changes requested):
After some testing and discussion there is now just one change I would still like to see before this gets merged: That is for the tests to check the post migration count of `PackageReleaseComponent` content.

Once that is added (hopefully not too difficult), my logic is that any amount of test suite, is a very big improvement over "no test suite whatsoever"!

quba42 (approved):
As far as I am concerned this is good to go. :+1:

goosemania (approved):

m-bucher:
Created a cherry-pick for 0.11.z in https://github.com/pulp/pulp-2to3-migration/pull/423

Do we also need a cherry-pick into 0.12.z?

goosemania:
> Created a cherry-pick for 0.11.z in #423
> 
> Do we also need a cherry-pick into 0.12.z?

@m-bucher, it's your choice. If you want to you can cherry-pick it there, no objections.
There are no policies at the moment that require to cherry-pick items to all releases up to the latest.
Somewhat inconvenient and surprising for users but that's the sad reality.

PR #403, opened 2021-07-01
Fix JSON serializable error

goosemania:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -149,18 +149,21 @@ def migrate_importers(plan):
                 pb.save()
 
 
-def complex_repo_migration(plugin, pulp3_repo_setup, repo_name):
+def complex_repo_migration(plugin_type, pulp3_repo_setup, repo_name):
     """Perform a complex migration for a particular repo using the repo setup config.
 
     Create all repository versions, publications, distributions.
 
     Args:
-        plugin: Plugin object
+        plugin_type(str): Plugin type
         pulp3_repo_setup: Pulp 3 repo setup config for a plugin
         repo_name: Name of the repo to be migrated
     """
-    distributor_migrators = plugin.migrator.distributor_migrators
-    distributor_types = list(plugin.migrator.distributor_migrators.keys())
+    from pulp_2to3_migration.app.plugin import PLUGIN_MIGRATORS
+    migrator = PLUGIN_MIGRATORS.get(plugin_type)
+
+    distributor_migrators = migrator.distributor_migrators
+    distributor_types = list(distributor_migrators.keys())
     repo_versions_setup = pulp3_repo_setup[repo_name]['repository_versions']
 
     # importer might not be migrated, e.g. config is empty or it's not specified in a MP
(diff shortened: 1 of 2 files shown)


pulp_2to3_migration/app/migration.py
```diff
@@ -283,7 +283,7 @@ def create_repoversions_publications_distributions(plan, parallel=True):
                 if needs_a_task:
                     repo_ver_to_create += len(repo_versions)
                     repo = Repository.objects.get(name=repo_name).cast()
-                    task_args = [plugin, pulp3_repo_setup, repo_name]
+                    task_args = [plugin.migrator.distributor_migrators, pulp3_repo_setup, repo_name]
```
goosemania, line 286: @dralley , @ggainey , so I found the spot what was not JSON serializable and a reason for CI to fail.
I thought I handled it well but ... the values of the `distributor_migrators` dict are class objects:
```
{'iso_distributor': pulp_2to3_migration.app.plugin.iso.repository.IsoDistributor}
```

Any ideas what to do?

The only thought I have is to pass a plugin name and then try to load all the configuration for using it. Not sure how easy it will be.

dralley: Well, there's always `eval()`.  It's safe so long as the user never controls the string being passed in, which should be the case here.

```python
distributor_migrators = {'iso_distributor': 'pulp_2to3_migration.app.plugin.iso.repository.IsoDistributor'}

## inside the task

distributor_migrators = {name: eval(classname) for name, classname in distributor_migrators.items()
```

ggainey: Hrm. I  think there's a more pythonic approach than eval(). Will dig around and see what I can find. Regardless of the specifics, tho, a core part of the new tasking system is "only simple types can be passed to tasks", so converting-from-string is prob the best we can do.

ggainey: Here's something that works (I'm not claiming any particular level of "good", just "works" :) ):
```python
import sys
clsstr = "pulp_2to3_migration.app.plugin.iso.repository.IsoDistributor"
modulename = ".".join(clsstr.split(".")[:-1])
classname = cstr.split(".")[-1]
d = getattr(sys.modules[modulename], classname)
print(d)
<class 'pulp_2to3_migration.app.plugin.iso.repository.IsoDistributor'>
```
I'm not sure this is better than eval() - other than being 'safer', and as dralley points out, we control what gets sent here.

ggainey: and here's a cleaner approach - which comes from django:

https://stackoverflow.com/a/34963527

pulp_2to3_migration/app/migration.py
```diff
@@ -149,18 +149,21 @@ def migrate_importers(plan):
...
     Args:
-        plugin: Plugin object
+        plugin_type(str): Plugin type
         pulp3_repo_setup: Pulp 3 repo setup config for a plugin
         repo_name: Name of the repo to be migrated
     """
-    distributor_migrators = plugin.migrator.distributor_migrators
-    distributor_types = list(plugin.migrator.distributor_migrators.keys())
+    from pulp_2to3_migration.app.plugin import PLUGIN_MIGRATORS
+    migrator = PLUGIN_MIGRATORS.get(plugin_type)
```
ggainey, line 163: :100:

pulp_2to3_migration/app/migration.py
goosemania: Thank you for looking into it!
Yeah, also the clsstr needs to be consrtuctred as well (from __module__ and __name__) :upside_down_face: 

I tried various ways, and found that passing a plugin type looks the least ugly to me, let know what you think.

ggainey (approved):
/me cheers wildly

PR #406, opened 2021-07-02
Update CI on 0.12 branch

goosemania:
[noissue]


goosemania:
I just realized that compatibility is needed for 3.14, so no Z release, it should be a Y one.

PR #412, opened 2021-07-19
Remove a test which attempts to check that only one reset can be run

goosemania:
The problem is that with the small amount of data which is used
for tests in CI, reset happens very quickly and very often is finished
before we are able to trigger the second task to test its failure.

The benefit of this test is low, the disturbance from regular test failures
because the first task is completed very quickly is high.

[noissue]


goosemania:
Alternatively, I can mark it as skipped but I feel like it will be skipped forever.

goosemania:
I suggest to merge it, current failures seem to be due to an attempt to use python 3.8, which has a proper configuration in CI in the other PR.

ipanova (approved):

dralley (approved):

PR #415, opened 2021-07-28
Fix modular errata migration

goosemania:
Wrong initialization of UpdateCollectionModule in createrepo_c which
led to all fields being set to null didn't cause any failures but
produced a wrong result `<module version=0/>`.

closes #8874
https://pulp.plan.io/issues/8874

### DIFF
--- a/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
+++ b/pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
@@ -410,7 +410,13 @@ class Pulp2Erratum(Pulp2to3Content):
             col.name = collection.get('name')
             module = collection.get('module')
             if module:
-                col.module = cr.UpdateCollectionModule(**module)
+                cr_module = cr.UpdateCollectionModule()
+                cr_module.name = module['name']
+                cr_module.stream = module['stream']
+                cr_module.version = int(module['version'])
+                cr_module.context = module['context']
+                cr_module.arch = module['arch']
+                col.module = cr_module
 
             for package in collection.get('packages', []):
                 pkg = cr.UpdateCollectionPackage()
--- a/CHANGES/8874.bugfix
+++ b/CHANGES/8874.bugfix
@@ -1,0 +1,1 @@
+Fixed modular errata migration.


pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
```diff
@@ -410,7 +410,13 @@ def create_pulp3_content(self):
...
             module = collection.get('module')
             if module:
-                col.module = cr.UpdateCollectionModule(**module)
+                cr_module = cr.UpdateCollectionModule()
+                cr_module.name = module.get('name')
+                cr_module.stream = module.get('stream')
+                cr_module.version = int(module.get('version', 0))
+                cr_module.context = module.get('context')
+                cr_module.arch = module.get('arch')
+                col.module = cr_module
```
dralley, line 419: \* **dalley** *cries profusely*

Yeah these constructors are c-style, they don't do anything with `args` or `kwargs`.  In `pulp_rpm`, we use the `to_createrepo_c` functions that avoid this problem by doing exactly this ^

This mistake is so easy to  overlook, I think we need to double-check everything :(

https://github.com/rpm-software-management/createrepo_c/blob/master/src/python/updatecollectionmodule-py.c#L90-L107

https://github.com/rpm-software-management/createrepo_c/blob/master/src/python/updaterecord-py.c#L95-L111

dralley (approved):

pulp_2to3_migration/app/plugin/rpm/pulp_2to3_models.py
dralley: Compare against https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/advisory.py#L304-L322, I notice a small difference with the module version - here it seems to be optional with a default, in `pulp_rpm` it is required.  I'm not sure which is correct off the top of my head

dralley: Luckily I believe this is the only place (at least in the migration plugin) we made this mistake, everything else looks correct.

goosemania: > Compare against https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/advisory.py#L304-L322, I notice a small difference with the module version - here it seems to be optional with a default, in `pulp_rpm` it is required. I'm not sure which is correct off the top of my head

It should be mandatory. thanks

goosemania: @dralley , I made it required. Thanks.
I do not use the `to_createrepo_c` method, since in the migration we approaching it from the other side. Create a createrepo_c object and make a pulp object from it. We do not have such use case in a plugin, we always either parse the xml or have pulp object to convert to createrepo_c.
Maybe we should have chosen to initialize pulp object first, create all the relations, and then convert it into the createrepo_c object. But I prefer not to refactor code here.

dralley: Makes sense, LGTM

PR #417, opened 2021-07-29
Update CI

goosemania:
[noissue]


.github/workflows/ci.yml
```diff
@@ -61,8 +61,6 @@ jobs:
       matrix:
         env:
           - TEST: pulp
-          - TEST: docs
-          - TEST: s3
```
goosemania, line 65: Why are those removed?

daviddavis: Because `test_s3` and `docs_test` are `false` in your template_config.yml. There was a bug that got fixed where these tests weren't being properly disabled.

ipanova (approved):

PR #420, opened 2021-07-30
Fix CI release job by updating /etc/pulp/settings.py

daviddavis:
Fixing the CI release test job which installs pulp-2to3-migration as a
package (instead of the project files). So we need to use
/etc/pulp/settings.py.

[noissue]

### DIFF
--- a/.github/workflows/scripts/post_before_script.sh
+++ b/.github/workflows/scripts/post_before_script.sh
@@ -2,8 +2,33 @@
 
 set -euv
 
-sudo sed -i  "s/bindIp: 127.0.0.1/bindIp: 127.0.0.1,$(ip address show dev docker0 | grep -o "inet [0-9]*\.[0-9]*\.[0-9]*\.[0-9]*" | grep -o "[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*")/g" /etc/mongod.conf
+export MONGODB_IP=$(ip address show dev docker0 | grep -o "inet [0-9]*\.[0-9]*\.[0-9]*\.[0-9]*" | grep -o "[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*")
+
+sudo sed -i  "s/bindIp: 127.0.0.1/bindIp: 127.0.0.1,$MONGODB_IP/g" /etc/mongod.conf
 sudo systemctl restart mongod
+
+# update settings to configure our mongo install
+cmd_prefix bash -c "cat >> /etc/pulp/settings.py <<EOF
+PULP2_MONGODB = {
+  'name': 'pulp_database',
+  'seeds': '$MONGODB_IP:27017',
+  'username': 'ci_cd',
+  'password': 'ci_cd',
+  'replica_set': '',
+  'ssl': False,
+  'ssl_keyfile': '',
+  'ssl_certfile': '',
+  'verify_ssl': True,
+  'ca_path': '/etc/pki/tls/certs/ca-bundle.crt',
+}
+EOF"
+
+# Restarting single container services
+cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-api"
+cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-content"
+cmd_prefix bash -c "s6-svc -r /var/run/s6/services/new-pulpcore-resource-manager"
+cmd_prefix bash -c "s6-svc -r /var/run/s6/services/new-pulpcore-worker@1"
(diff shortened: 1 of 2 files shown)


.github/workflows/scripts/post_before_script.sh
```diff
@@ -5,6 +5,23 @@ set -euv
 sudo sed -i  "s/bindIp: 127.0.0.1/bindIp: 127.0.0.1,$(ip address show dev docker0 | grep -o "inet [0-9]*\.[0-9]*\.[0-9]*\.[0-9]*" | grep -o "[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*")/g" /etc/mongod.conf
 sudo systemctl restart mongod
 
+# update settings to configure our mongo install
+export MONGODB_IP=$(ip address show dev docker0 | grep -o "inet [0-9]*\.[0-9]*\.[0-9]*\.[0-9]*" | grep -o "[0-9]*\.[0-9]*\.[0-9]*\.[0-9]*")
```
goosemania, line 9: We can maybe do this on line 4 to use $MONGO_IP on line 5 as well?

daviddavis:
This won't work as by the time we update the settings, it's too late and pulp has already started with the old settings.

My initial thought was that we could set the settings [via the template_config.yml](https://git.io/JBVPJ) but we need to dynamically set the hostname.

I need to think of another solution that would work.

daviddavis:
@pulp/continuous-integration if you have any thoughts ^^^

fao89:
> @pulp/continuous-integration if you have any thoughts ^^^

I don't understand the issue, why do we need to change the settings?

daviddavis:
> I don't understand the issue, why do we need to change the settings?

We have to configure pulp in our CI to connect to mongo (see the change in this PR). The ip address needs to be found dynamically though so using `pulp_settings` in template_config.yml I don't think is an option.

daviddavis:
If you're asking about what is prompting this change: it's because the release process is broken. Currently, the CI code updates the code in pulp-2to3-migration git repo which gets mounted into the pulp container and changes the settings in the pulp container:

https://github.com/pulp/pulp-2to3-migration/blob/master/.github/workflows/scripts/post_before_install.sh

However, now the release project generates a pulp-2to3-migration package and tests against that instead of using the git repo checkout.

fao89:
> If you're asking about what is prompting this change: it's because the release process is broken. Currently, the CI code updates the code in pulp-2to3-migration git repo which gets mounted into the pulp container and changes the settings in the pulp container:
> 
> https://github.com/pulp/pulp-2to3-migration/blob/master/.github/workflows/scripts/post_before_install.sh
> 
> However, now the release project generates a pulp-2to3-migration package and tests against that instead of using the git repo checkout.

Oh! Now I can understand the problem, thank you!

fao89:
> This won't work as by the time we update the settings, it's too late and pulp has already started with the old settings.
> 
> My initial thought was that we could set the settings [via the template_config.yml](https://git.io/JBVPJ) but we need to dynamically set the hostname.
> 
> I need to think of another solution that would work.

you can restart pulp-api service:
https://github.com/pulp/pulpcore/blob/master/.github/workflows/scripts/script.sh#L128-L139

.github/workflows/scripts/post_before_script.sh
```diff
@@ -2,9 +2,37 @@
...
+
+# Restarting single container services
+cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-api"
+cmd_prefix bash -c "s6-svc -r /var/run/s6/services/pulpcore-content"
+cmd_prefix bash -c "s6-svc -d /var/run/s6/services/pulpcore-resource-manager"
+cmd_prefix bash -c "s6-svc -d /var/run/s6/services/pulpcore-worker@1"
+cmd_prefix bash -c "s6-svc -d /var/run/s6/services/pulpcore-worker@2"
+cmd_prefix bash -c "s6-svc -u /var/run/s6/services/new-pulpcore-resource-manager"
+cmd_prefix bash -c "s6-svc -u /var/run/s6/services/new-pulpcore-worker@1"
+cmd_prefix bash -c "s6-svc -u /var/run/s6/services/new-pulpcore-worker@2"
```
fao89, line 34: -d is stopping old tasking system, -u is starting the new tasking system
https://skarnet.org/software/s6/s6-svc.html
it makes sense for the upgrade test, but for this I believe you need -r for the new tasking system

fao89 (approved):

goosemania (approved):

daviddavis:
@goosemania note that if you backport this change to release branches you’ll have to tweak the lines of code that restart the workers to use the old style workers.

PR #434, opened 2021-08-23
Runs all DB queries synchronously during a sync task

fao89:
Required PR: https://github.com/pulp/pulpcore/pull/1553
closes #9283


dralley:
@goosemania How long do we plan on supporting the migration plugin for mainline pulpcore?

PR #438, opened 2021-08-27
Fix the case when pulp2 importer is not migrated but its LCE is used.

goosemania:
closes #8862
https://pulp.plan.io/issues/8862

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -305,8 +305,10 @@ class ContentMigrationFirstStage(Stage):
                     base_path = pulp2content.pulp2_storage_path
                     remotes = set()
                     missing_artifact = False
+                    remote_declarative_artifacts = []
 
                     for image_relative_path in extra_info['download']['images']:
+                        remote_url_tuples = []
                         image_path = os.path.join(base_path, image_relative_path)
                         downloaded = os.path.exists(image_path)
                         if downloaded:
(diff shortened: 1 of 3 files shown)


goosemania:
There are situations when we look up a remote by its pulp2importer counter part, and we get `None`. This needs to be ignored (for downloaded content) or skipped with a warning (for on_demand content). 

There are changes in 2 parts of content.py (which is the most tricky and unreadable part of this whole subproject).
The changes are very similar in the approach.
The choice I think we have is to:
 - introduce more nested if/else's with duplicated code
 - OR try to collect needed data related to remotes first, filter out missed ones and then use this info to create declarative artifacts.
 
This PR attempts to implement the latter.

goosemania:
@dralley , @ggainey , if you have an opinion on the chosen approach, or have any suggestions, they are very welcome.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -416,46 +425,46 @@ def get_remote_by_importer_id(importer_id):
...
                                 deferred_download=deferred_download)
                             lce.is_migrated = True
-                            at_least_one_lce_migrated = True
                             dc = DeclarativeContent(content=pulp3content, d_artifacts=[da])
+                            future_relations.update({'lces': list(pulp2lazycatalog)})
                             dc.extra_data = future_relations
                             await self.put(dc)
 
-                        if not at_least_one_lce_migrated:
+                        if not remote_lce_tuples and deferred_download:
```
goosemania, line 459: Note for myself. Need to cover the case when content has LCE but no related importers are migrated, and content is downloaded, so we can actually proceed. In this implementation it is skipped.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -320,30 +322,35 @@ def get_remote_by_importer_id(importer_id):
                             artifact = Artifact()
 
                         lces = pulp2lazycatalog.filter(pulp2_storage_path=image_path)
-                        if lces:
-                            remote_declarative_artifacts = []
 
+                        if not lces and not downloaded:
+                            continue
+
+                        if lces:
```
ipanova, line 329: this condition is unnecessary

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -416,46 +425,49 @@ def get_remote_by_importer_id(importer_id):
...
                             da = DeclarativeArtifact(
                                 artifact=artifact,
                                 url=lce.pulp2_url,
                                 relative_path=relative_path,
                                 remote=remote,
                                 deferred_download=deferred_download)
                             lce.is_migrated = True
-                            at_least_one_lce_migrated = True
                             dc = DeclarativeContent(content=pulp3content, d_artifacts=[da])
+                            future_relations.update({'lces': list(pulp2lazycatalog)})
```
ipanova, line 457: i think this update should happen  outside the for loop

pulp_2to3_migration/app/plugin/content.py
goosemania: Could you elaborate?

pulp_2to3_migration/app/plugin/content.py
goosemania: It used to be and I can't get my head around why.
It might be a bit redundant but at least LCE will be updated.
When it is outside of the loop, I'm not sure how correct LCEs go to the next stage if we are assigning them to the future_relations after we pushed content to the queue for the next stage.

pulp_2to3_migration/app/plugin/content.py
ipanova: You can just remove the if condition and keep only the for loop. the `lces` from line 324 will wether be an empty queryset or will contain some data to loop trough.

pulp_2to3_migration/app/plugin/content.py
ipanova: after a short call we had, it was decided to leave these changes as they are and leave a comment in the code. The confusion was around assigning in the future_relations all lazy catalog to the dc instead of one single lce.

ipanova (approved):

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -416,46 +424,55 @@ def get_remote_by_importer_id(importer_id):
...
                                 artifact=artifact,
                                 url=lce.pulp2_url,
                                 relative_path=relative_path,
                                 remote=remote,
                                 deferred_download=deferred_download)
                             lce.is_migrated = True
-                            at_least_one_lce_migrated = True
                             dc = DeclarativeContent(content=pulp3content, d_artifacts=[da])
+
+                            # yes, all LCEs are assigned for each dc to be resolved at a later
```
ipanova, line 457: :+1:

goosemania:
Re-opened against master https://github.com/pulp/pulp-2to3-migration/pull/451.
Multiple z-stream releases do not let to change the base easily.

goosemania:
Reopening to release the 0.11.z

PR #441, opened 2021-08-31
Give in to the double quotes in setup.py

goosemania:
[noissue]


goosemania:
This is so all release automation works fine, including "create release branch workflow".

PR #451, opened 2021-09-10
Fix the case when pulp2 importer is not migrated but its LCE is used.

goosemania:
closes #8862
https://pulp.plan.io/issues/8862


goosemania:
Against master

ipanova (approved):

PR #462, opened 2021-10-11
Fix compatibility with pulpcore 3.16

goosemania:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/plugin/content.py
+++ b/pulp_2to3_migration/app/plugin/content.py
@@ -6,6 +6,8 @@ import shutil
 
 from gettext import gettext as _
 
+from asgiref.sync import sync_to_async
+
 from django.conf import settings
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import transaction
@@ -29,12 +31,14 @@ from pulpcore.plugin.stages import (
     RemoteArtifactSaver,
     Stage,
 )
+from pulpcore.plugin.sync import sync_to_async_iterable
 
 from pulp_2to3_migration.app.constants import (
     DEFAULT_BATCH_SIZE,
     NOT_USED,
 )
 from pulp_2to3_migration.app.models import (
+    Pulp2Content,
     Pulp2Importer,
     Pulp2LazyCatalog,
 )
@@ -235,7 +239,7 @@ class ContentMigrationFirstStage(Stage):
         async with ProgressReport(
             message='Migrating {} content to Pulp 3'.format(content_type),
             code='migrating.{}.content'.format(self.migrator.pulp2_plugin),
-            total=pulp_2to3_detail_qs.count()
+            total=await sync_to_async(pulp_2to3_detail_qs.count)()
         ) as pb:
             select_extra = [
                 'pulp2content',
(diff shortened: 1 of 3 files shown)


goosemania:
@dralley , @dkliban , @ipanova , CI is green but the code in `content.py` is ugly.
I would prefer to move large pieces into separate functions and just call them wrapped in `sync_to_async`, but the current code is a mixture of sync and async stuff, so it seems to be hard to do that without major rewrite/refactor.
Thoughts or suggestions?

dkliban (approved):

dralley:
>I would prefer to move large pieces into separate functions and just call them wrapped in sync_to_async, but the current code is a mixture of sync and async stuff, so it seems to be hard to do that without major rewrite/refactor.

Yeah, I don't think we should spend any more time on it than we absolutely need to.  I'm fine with not doing any substantial refactors.

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -246,16 +249,21 @@ def get_remote_by_importer_id(importer_id):
                 select_extra.append('pulp2content__pulp2_repo')
 
             pulp_2to3_detail_qs = pulp_2to3_detail_qs.select_related(*select_extra)
-            for pulp_2to3_detail_content in pulp_2to3_detail_qs.iterator(chunk_size=800):
+            async for pulp_2to3_detail_content in sync_to_async_iterable(
+                    pulp_2to3_detail_qs.iterator(chunk_size=800)
+            ):
                 dc = None
                 pulp2content = pulp_2to3_detail_content.pulp2content
+                await sync_to_async(pulp2content.refresh_from_db)()  # force model instance to be
```
dralley, line 257: What's going on here?

goosemania: pulp2content is used in many places later, this is to force to load fully from the db without a need to do db requests later when different fields are accessed.

ipanova:
> @dralley , @dkliban , @ipanova , CI is green but the code in `content.py` is ugly. I would prefer to move large pieces into separate functions and just call them wrapped in `sync_to_async`, but the current code is a mixture of sync and async stuff, so it seems to be hard to do that without major rewrite/refactor. Thoughts or suggestions?

I'd leave this is as is. The changes are easily readable.

pulp_2to3_migration/app/plugin/content.py
ipanova: I think whether get() or refresh_from_db() is used there is not much of the difference. refresh_from_db() calls anyway into get() its just does not load deferred fields and clears cached relations https://github.com/django/django/blob/main/django/db/models/base.py#L652

ipanova (approved):

pulp_2to3_migration/app/plugin/content.py
```diff
@@ -246,16 +250,22 @@ def get_remote_by_importer_id(importer_id):
...
 
             pulp_2to3_detail_qs = pulp_2to3_detail_qs.select_related(*select_extra)
-            for pulp_2to3_detail_content in pulp_2to3_detail_qs.iterator(chunk_size=800):
+            async for pulp_2to3_detail_content in sync_to_async_iterable(
+                    pulp_2to3_detail_qs.iterator(chunk_size=800)
+            ):
                 dc = None
-                pulp2content = pulp_2to3_detail_content.pulp2content
+                pulp2content = await sync_to_async(Pulp2Content.objects.get)(
+                    pulp_2to3_detail_content.pulp2content_id
```
dralley, line 258: I think it needs `pk=`, my bad

goosemania: yeah, I just got to it and figured that, we are perfectly in sync :)

PR #464, opened 2021-10-18
Add default signing-service to complex plan

m-bucher:
closes #9517
https://pulp.plan.io/issues/9517

### DIFF
--- a/pulp_2to3_migration/app/migration.py
+++ b/pulp_2to3_migration/app/migration.py
@@ -11,6 +11,7 @@ from pulpcore.plugin.models import (
     CreatedResource,
     ProgressReport,
     Repository,
+    SigningService,
     TaskGroup,
 )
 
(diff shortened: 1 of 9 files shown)


pulp_2to3_migration/app/migration.py
```diff
@@ -166,6 +167,20 @@ def complex_repo_migration(plugin_type, pulp3_repo_setup, repo_name):
...
+        _logger.info(
+            "Signing Service %r requested for %r", signing_service_name, repo_name
+        )
+        try:
+            signing_service = SigningService.objects.get(name=signing_service_name)
+        except SigningService.DoesNotExist:
+            _logger.warning(
+                "Could not find signing-service named %r", signing_service_name
+            )
+            raise
```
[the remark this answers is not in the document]
goosemania: I think it's fine. I would fail since you explicitly specify in the migration plan that a certain signing service needs to be used.

I hoped that it would be a more local, plugin specific change. But since we do not provide any hooks to migrate a pulp2 repo in a custom way, the only option would be to have this check for signing service in a distributor migration. And this would mean a lot of duplicated checks for signing service existence (N repo versions and for each you use the same signing service). For that reason, I'm ok with this code to be here, though not entirely desirable.

goosemania:
FWIW, CI is fixed, consider rebasing if you see unrelated test failures.

CHANGES/9517.misc
```diff
@@ -0,0 +1 @@
+Added ``signing_service`` for (apt-)repository in Complex-MigrationPlan.
```
goosemania, line 1: I would consider this a feature, rather than just a misc.
User defined migration plan is being extended with this change.

Somehting like `Added an ability to use a manually pre-created Pulp 3 SigningService during pulp_deb repository migration. A migration plan now accepts ``signing_service`` key for each repository.`

What do you think?

goosemania (commented):
Could you change the changelog type and expand a bit on the use of the change? See my comment. It looks good otherwise. Thanks!

quba42:
@goosemania Just FYI: We are personally not using the migration plugin past the `0.11` branch (since our use case is Katello), so that is also where we will be back porting this change to. You can now make of this information what you will. :wink: 

Your review for this change is obviously valuable to us (no matter what branch it goes into).

CHANGES/9517.misc
m-bucher: Sounds good, done :smiley:

goosemania (approved):

PR #466, opened 2021-11-11
Add DEB_COMPONENT_BATCH_SIZE setting

quba42:
closes #9564
https://pulp.plan.io/issues/9564

Since each deb_components creates a large number of Pulp2to3Content we
need a much lower batch size for this type.

### DIFF
--- a/pulp_2to3_migration/app/settings.py
+++ b/pulp_2to3_migration/app/settings.py
@@ -14,3 +14,7 @@ PULP2_MONGODB = {
 ALLOWED_CONTENT_CHECKSUMS = ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
 
 CONTENT_PREMIGRATION_BATCH_SIZE = 1000
+
+# Since each deb_component creates a large number of Pulp2to3Content we need a much lower batch size
+# for this type, in order to avoid CursorNotFound errors!
+DEB_COMPONENT_BATCH_SIZE = 50
--- a/CHANGES/9564.bugfix
+++ b/CHANGES/9564.bugfix
@@ -1,0 +1,1 @@
+deb_components now use a default batch size of 50 to avoid CursorNotFound errors during pre_migration for this one to many type.
--- a/CHANGES/9564.feature
+++ b/CHANGES/9564.feature
@@ -1,0 +1,1 @@
+Added the `DEB_COMPONENT_BATCH_SIZE` setting, so users can individually control the batch size for this one to many type.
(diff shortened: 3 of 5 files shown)


quba42:
Note that I have not yet had a chance to verify the fix, I will post again when I do.

pulp_2to3_migration/app/settings.py
```diff
@@ -14,3 +14,7 @@
 ALLOWED_CONTENT_CHECKSUMS = ['md5', 'sha1', 'sha224', 'sha256', 'sha384', 'sha512']
 
 CONTENT_PREMIGRATION_BATCH_SIZE = 1000
+
+# Since each deb_component creates a large number of Pulp2to3Content we need a much lower batch size
+# for this type, in order to avoid CursorNotFound errors!
+DEB_COMPONENT_BATCH_SIZE = 50
```
goosemania, line 20: Can the `CONTENT_PREMIGRATION_BATCH_SIZE` be configured for the value needed?
What is the reason to have a debian specific config for that?

There are multiple cases when the default value for the batch size needs a change. It can be slow io, low memory, many old errata (rpm plugin), etc.
I understand that with having a plugin specific batch size, there is more flexibility and other plugin migrations are not affected by the change but it doesn't seem to make much of a difference if there is at least one reason to make it lower. The overall migration will be somewhat slower but it's in the background. 

Let me know if I miss anything.
Imo, adding a plugin specific config and code to the general migration code base should have a strong reason and not much way around it.

quba42: This setting should affect only the deb_component content type. The other pulp_deb types are also unaffected, and use the normal batch size instead. The deb_component type has a one to many relationship in how many Pulp 3 units are created. That is, each deb_component in Pulp 2 might create hundreds, thousands, or in the case of Debian and Ubuntu "main" components, tens of thousands of Pulp3 units (as well as the relevant intermediary 2to3 types). As a result, even if the default batch size of let's say 1000 is appropriate for every other content type, deb_component might need to use a significantly lower batch size in order to avoid `CursorNotFound` errors.

One alternative to making the batch size freely configurable for deb_components, might be to scale it down by a fixed factor, perhaps `batch_size / 100`. However that would be a lot less flexible, and it is difficult for me to decide what the fixed factor should be for all systems out there...

quba42:
@goosemania Ok, I have now finally managed to confirm on a test system that these configuration values including there defaults all work as expected. Ready for final review/merge.

quba42:
Tests appear to be failing because the CI is still looking for `master` branch of the container plugin (which was presumably renamed to `main`).

goosemania:
Looks good, thanks!
Please rebase, I fixed the CI, it turned out to be more than just branches at the end :)

goosemania (approved):

PR #472, opened 2021-11-17
Apply the latest plugin template to the 0.11 branch

goosemania:
[noissue]


template_config.yml
```diff
@@ -22,38 +22,49 @@ check_gettext: true
 check_manifest: true
 check_openapi_schema: true
```
fao89, line 23: you have to disable it,
OpenAPI was fixed at pulpcore 3.10
https://github.com/pulp/pulpcore/pull/1065

PR #474, opened 2021-12-02
Guard pulp3_remote from importer clearing

quba42:
closes #8968
https://pulp.plan.io/issues/8968

We have hit the above issue while 2to3 migrating SLES importers.
Apparently this is sometimes called for SLES importers that already have
their pulp3_remote set to None, resulting in:

'NoneType' object has no attribute 'delete'

We did not perform any deeper analysis of why this is the case, but
simply added this guard to fix the problem.

### DIFF
--- a/CHANGES/8968.bugfix
+++ b/CHANGES/8968.bugfix
@@ -1,0 +1,1 @@
+Fixed ``'NoneType' object has no attribute 'delete'`` error during migration re-runs.
--- a/pulp_2to3_migration/app/pre_migration.py
+++ b/pulp_2to3_migration/app/pre_migration.py
@@ -605,7 +605,8 @@ def pre_migrate_importer(repo_id, importer_types):
         if last_updated != importer.pulp2_last_updated:
             # remove Remote in case of feed change
             if importer.pulp2_config.get('feed') != importer_data.config.get('feed'):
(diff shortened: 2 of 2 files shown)


quba42:
@goosemania I opened this PR more as a FYI. We added this guard on our internal 0.11.5 build, since we sometimes hit this issue while migrating SLES remotes. I did not perform much analysis, I simply added the guard which fixed the issue for us...

goosemania:
It makes sense and it's a safe thing to do, so it looks good, I'd merge it.

I can imagine a case when something when wrong with the first migration attempt (or it was interrupted), so some importers were not migrated, and by the next migration run importer's feed changed in pulp2.  That's the only scenario I can see when it can happen. And your solution is the only thing which can be and should be done, imo.

CHANGES/8968.bugfix
```diff
@@ -0,0 +1 @@
+Added a guard to avoid a rare ``'NoneType' object has no attribute 'delete'`` error.
```
goosemania, line 1: Could it be rephrased, to avoid the word guard? E.g. 
```
Fixed ``'NoneType' object has no attribute 'delete'`` error during migration re-runs.
```

goosemania (approved):

PR #477, opened 2021-12-07
[cp] Guard pulp3_remote from importer clearing

goosemania:
backports #8968
https://pulp.plan.io/issues/8968

We have hit the above issue while 2to3 migrating SLES importers.
Apparently this is sometimes called for SLES importers that already have
their pulp3_remote set to None, resulting in:

'NoneType' object has no attribute 'delete'

We did not perform any deeper analysis of why this is the case, but
simply added this guard to fix the problem.

fixes #9612

(cherry picked from commit b8892d7d15317b5e46ef568bc6c2ccbddacb46b0)

### DIFF
--- a/.github/workflows/scripts/install.sh
+++ b/.github/workflows/scripts/install.sh
@@ -16,6 +16,7 @@ set -euv
 source .github/workflows/scripts/utils.sh
 
 if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
+  pip install "mistune<2.0.0"
   pip install -r ../pulpcore/doc_requirements.txt
   pip install -r doc_requirements.txt
 fi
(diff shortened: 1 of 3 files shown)


.ci/ansible/Containerfile.j2
```diff
@@ -21,6 +21,9 @@ RUN pip3 install \
 {%- if item.name == "pulp-certguard" -%}
 {{ " " }}python-dateutil rhsm
 {%- endif -%}
+{%- if item.name == "pulp-2to3-migration" -%}
```
fao89, line 24: if it doesn't work, try:
{%- if item.name == "pulpcore" -%}

goosemania: Container building part was too late but you set me to look into the right direction, thanks!

PR #478, opened 2021-12-08
Ensure deb uniqueness constraints allow for one 2to3 unit per Pulp2 unit

quba42:
closes #9616
https://pulp.plan.io/issues/9616

Previously the uniqueness constraints allowed for only one 2to3 unit per
Pulp 3 unit to be created. This causes problems when there is more than
one Pulp 2 unit that map onto the same Pulp 3 unit to be created.


pulp_2to3_migration/app/migrations/0031_add_repoid_to_deb_types.py
```diff
@@ -0,0 +1,53 @@
...
+
+    dependencies = [
+        ('pulp_2to3_migration', '0030_update_json_field'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='pulp2debcomponent',
+            name='repoid',
+            field=models.TextField(default='blub'),
```
quba42, line 16: For now, I have not tried to implement anything to add a correct `repoid` to already created entries.
To be honest I feel like in this case with the migration plugin it makes more sense to drop all entries from the affected tables and then re-run the deb migration...

Just setting `default='blub'` is clearly not a real solution, but I thought I would ask if there is any experience with how to handle such cases before implementing something random. The current migration only works if it is applied before the relevant tables have any entries.

quba42 (commented):
Our test system that reproduces the issue is currently still migrating, but we have good reason to believe it will work, since a prior iteration of the fix was partially successful. I will post again when this latest test run is complete.

Which just leaves the DB migration issue described in the comment below.

quba42:
Test migration was successful.

goosemania:
I agree that users need to run the reset command before attempting a new migration run after applying this fix.

Alternatively you can just add `pulp2content` to the uniqueness constraint and do not bother with  repo_id if you are not using it, up to you. DB migration wise, handling repo_id is easier.

pulp_2to3_migration/app/migrations/0031_add_repoid_to_deb_types.py
goosemania: I think it's fine to set repo_id to some string for the db migration sake.
You can't fix db data properly for users to just rerun a content migration (maybe in theory you can but it's difficult and error prone). They will need to run reset first anyway, so your random string in the repo_id will be cleaned up during reset.

quba42:
> I agree that users need to run the reset command before attempting a new migration run after applying this fix.
> 
> Alternatively you can just add `pulp2content` to the uniqueness constraint and do not bother with repo_id if you are not using it, up to you. DB migration wise, handling repo_id is easier.

I can see it working equally well with `pulp2content`, but given it is working now, I am reluctant to change it unless there is reason to believe it will work better than before. It's true that I am not actually using the repo_id other than to mimic the Pulp 2 uniqueness constraints. Do you think relating the `pulp2content` would be more efficient?

goosemania:
> > I agree that users need to run the reset command before attempting a new migration run after applying this fix.
> > Alternatively you can just add `pulp2content` to the uniqueness constraint and do not bother with repo_id if you are not using it, up to you. DB migration wise, handling repo_id is easier.
> 
> I can see it working equally well with `pulp2content`, but given it is working now, I am reluctant to change it unless there is reason to believe it will work better than before. It's true that I am not actually using the repo_id other than to mimic the Pulp 2 uniqueness constraints. Do you think relating the `pulp2content` would be more efficient?

Nah, I think you are good.

quba42:
Squished the two commits since I think this is as ready as it will be.

quba42:
I can update the issue reference once https://github.com/pulp/pulp-2to3-migration/pull/485 is merged.

goosemania:
The CI update is merged, please update the issue number in the commit and in the CHANGES to the number of the issue on github.

goosemania (approved):

PR #480, opened 2021-12-09
Use better deb subid type names

quba42:
[noissue]

Since these dict keys are user facing via migration progress reports
(but have no other impact) it would be nice to have meaningful names.


quba42:
I will mark this as "Ready for review" once I have seen some internal test migrations using it.
That being said I am 99% sure these dict keys have no other impact other than as a label in progress report messages.

goosemania (approved):

PR #485, opened 2021-12-23
Update CI after the move to github issues.

goosemania:
[noissue]


fao89 (approved):
LGTM!

You can also delete:
- https://github.com/pulp/pulp-2to3-migration/blob/main/.ci/scripts/redmine.py
- https://github.com/pulp/pulp-2to3-migration/blob/main/.ci/scripts/update_redmine.sh

PR #505, opened 2022-01-17
Only migrate the last 10 changelogs for each package

dralley:
[noissue]

### DIFF
--- a/pulp_2to3_migration/app/plugin/rpm/xml_utils.py
+++ b/pulp_2to3_migration/app/plugin/rpm/xml_utils.py
@@ -234,6 +234,8 @@ def parse_repodata(primary_xml, filelists_xml, other_xml):
     cr.xml_parse_primary_snippet(primary_xml, pkgcb=pkgcb, do_files=False)
     cr.xml_parse_filelists_snippet(filelists_xml, newpkgcb=newpkgcb)
     cr.xml_parse_other_snippet(other_xml, newpkgcb=newpkgcb)
+    # keep only the last 10 changelogs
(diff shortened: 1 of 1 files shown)


dralley:
In our previous discussions I had not really been enthused about making changes here, but since the implementation is *so* simple, and since we've been getting complaints about how long the migrations take (not specific to this but just generally), maybe it's still a good idea?

Even if we don't backport it to the downstream branch it could make sense to keep around upstream.

pulp_2to3_migration/app/plugin/rpm/xml_utils.py
```diff
@@ -234,6 +234,8 @@ def newpkgcb(pkgId, name, arch):
     cr.xml_parse_primary_snippet(primary_xml, pkgcb=pkgcb, do_files=False)
     cr.xml_parse_filelists_snippet(filelists_xml, newpkgcb=newpkgcb)
     cr.xml_parse_other_snippet(other_xml, newpkgcb=newpkgcb)
+    # keep only the last 10 changelogs
+    package.changelogs = package.changelogs[-10:]
```
dralley, line 238: This could be more like `-getattr("KEEP_CHANGELOG_LIMIT", settings, 1))`

goosemania: I think for simplicity let's keep it hard-coded until anyone asks to change it.

goosemania: Also this way, it can be hard-coded if needed.

goosemania (approved):

PR #516, opened 2022-03-09
Make DistributionTree migration compatible with pulp_rpm 3.17+.

### LINKED ISSUE #491: Alma Linux 8 or CentOS 8 repos migration can fail with: No declared artifact with relative path "images/boot.iso" (or ".treeinfo")
See details here https://github.com/pulp/pulp_rpm/issues/2305.

pulp-2to3-migration needs to take provide digest during DistributionTree creation.
This fix will be a compatibility release with pulp_rpm containing a fix for https://github.com/pulp/pulp_rpm/issues/2305.

goosemania:
closes #491

### DIFF
--- a/CHANGES/491.bugfix
+++ b/CHANGES/491.bugfix
@@ -1,0 +1,4 @@
+Fixed DistributionTree (kickstart tree) migration for pulp_rpm 3.17+.
+
+Alma Linux 8 or CentOS 8 repos migration will no longer fail with:
+No declared artifact with relative path "images/boot.iso" (or ".treeinfo").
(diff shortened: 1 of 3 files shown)


dralley (approved):

goosemania:
I'm not sure which pulpcore version to require. This will work with pulp_rpm 3.17+.
We can specify a requirement only for pulpcore. I'm thinking to require pulpcore>=3.18, even though 3.18 is not necessarily needed but this way we'll exclude older version of pulp_rpm.
Thoughts? @ggainey @dralley

dralley:
Given the cirucmstances `>=3.18` probably makes sense. 

Now that plugins can be removed, should we consider setting the EOL for this plugin?  (apart from things we need to backport, of course)

goosemania:
I would like to find time to add a test to prove that removal works and then +1 to set EOL. How does that sound?

dralley:
+1

PR #520, opened 2022-03-14
Use underscore twice to create anonymous links for CHANGES.rst

goosemania:
Since we no longer file an issue for each backport, the same
issue number is present in multiple releases. When there is
more than one link with the same name, sphinx throws a warning
`Duplicate explicit target name: #<github issue number>`.
To avoid that use two underscores to make an anonymous link.

[noissue]

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ filename = "CHANGES.rst"
 directory = "CHANGES/"
 title_format = "{version} ({project_date})"
 template = "CHANGES/.TEMPLATE.rst"
-issue_format = "`#{issue} <https://github.com/pulp/pulp-2to3-migration/issues/{issue}>`_"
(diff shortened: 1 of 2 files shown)


goosemania:
This failure https://github.com/pulp/pulp-2to3-migration/runs/5516778079?check_suite_focus=true#step:16:105 led me here

pyproject.toml
```diff
@@ -11,7 +11,7 @@ filename = "CHANGES.rst"
 directory = "CHANGES/"
 title_format = "{version} ({project_date})"
 template = "CHANGES/.TEMPLATE.rst"
-issue_format = "`#{issue} <https://github.com/pulp/pulp-2to3-migration/issues/{issue}>`_"
+issue_format = "`#{issue} <https://github.com/pulp/pulp-2to3-migration/issues/{issue}>`__"
```
fao89, line 14: 👍🏽

fao89 (approved):

PR #549, opened 2022-05-31
Fix call to remove_content() with QuerySet object

phess:
This PR was originally sent as a pulpcore change but that was misplaced in terms of code.

This PR fixes a situation where the migration plugin calls `remove_content()` and passes it a `Content` object instead of a `QuerySet`. The pulpcore `repository.py` code then checks for the `count()` method for the passed object which _is_ implemented for `QuerySet` but not for `Content`.

So passing a `Content` object causes pulpcore's `repository.py`'s `remove_content()` to fail due to the missing `count()` method.
The fix I'm implementing here is to build a `QuerySet` out of this single `Content` object so as to satisfy pulpcore.

### DIFF
--- a/CHANGES/550.bugfix
+++ b/CHANGES/550.bugfix
@@ -1,0 +1,6 @@
+Fixed issue where migration.py passes a Content object to remove_content(),
+which ends up breaking pulpcore's remove_content() further down the line with
+a traceback stating, 'Content' object has no attribute 'count'.
+
+Pulpcore's remove_content() will not face this issue anymore as it will now
+always receive a QuerySet object from migration.py.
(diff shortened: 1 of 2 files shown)


dralley (approved):

ipanova:
The PR is looking good, thank you!
can open an issue https://github.com/pulp/pulp-2to3-migration/issues and in the commit message reference this issue in the format of `closes #1234`
It also needs a changelog entry since this is a bug fix. Please take a  look at this PR as an example https://github.com/pulp/pulp-2to3-migration/pull/516/commits

ipanova:
@phess i guess you want this backported to 0.11?

phess:
> @phess i guess you want this backported to 0.11?

Yes please, if it's sufficiently trivial to backport.

Thank you so much for asking.

PR #552, opened 2022-06-02
Update CI files

pulpbot:
[noissue]


dralley:
@ipanova It looks like the nightly CI started failing around the same time, so I don't think this PR is failing for any reason specific to the PR.

ipanova:
@dralley I agree that the failure is not specific to this branch/PR. I do observe same traceback on other branches too. Thanks for looking into this.

dralley (approved):

PR #564, opened 2022-06-27
Update CI

fao89:


fao89:
https://github.com/pulp/pulp-2to3-migration/pull/563

PR #589, opened 2022-09-20
Workaround for releasing

fao89:
[noissue]


ipanova (approved):

fao89:
@dkliban what was the solution you found for pulp_ansible? 
We are hitting the same issue here (the test with the settings file)

PR #600, opened 2022-11-04
Fix CI by staying on ubuntu-20.04, which has mongo

mikedep333:
[noissue]


ggainey (approved):
great, just what we needed

